PR #1, opened 2015-10-23
Create development VM

arothuis:
Part of [106189596](https://www.pivotaltracker.com/story/show/106189596)

Borrowed from the `conext-operation-support` dev vm.


Vagrantfile
```diff
@@ -0,0 +1,56 @@
...
+
+Vagrant.configure("2") do |config|
+    config.vm.provider :virtualbox do |vm|
+        vm.name = "OpenConext Profile"
+        vm.memory = 1024
+    end
+
+    config.vm.box = "jayunit100/centos7"
+
+    config.vm.hostname = "profile-dev.engine.coin.surf.net"
```
DRvanR: use profile.vm.openconext.org, makes it easier to port back to openconext-deploy in the future.

ansible/group_vars/all.yml
```diff
@@ -0,0 +1,19 @@
+---
+user: profile
+timezone: Europe/Amsterdam
+server:
+  name: profile-dev.engine.coin.surf.net
```
DRvanR: same here ;)

Vagrantfile
arothuis: Makes sense :)

ansible/roles/app/tasks/main.yml
```diff
@@ -0,0 +1,8 @@
+- name: Test if app_dev.php.dist exists
+  command: test -f /var/www/{{ server.name }}/app_dev.php.dist
```
DRvanR: please place this file inside that ansible/roles/app/files folder, as that is where it belongs - and only copy it when in dev mode :tongue:

Vagrantfile
thijskh: Slightly tangential: we try to avoid references to "SURF" or "SURFnet" in the OpenConext repositories as (in theory) SURFconext is just one installation of this software.

arothuis:
@DRvanR @rjkip

ansible/roles/app/files/app_dev.php.dist
```diff
@@ -0,0 +1,24 @@
+<?php
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Debug\Debug;
+
+if (isset($_SERVER['HTTP_CLIENT_IP'])
```
DRvanR: please remove this check completely, as we can't do a thing with it in place :wink:

ansible/development.yml
```diff
@@ -0,0 +1,10 @@
+---
+- hosts: all
+  sudo: yes
+  roles:
+    - common
+    - rsyslog
+    - nginx
+    - mariadb
```
DRvanR: do we need Maria?

ansible/roles/nginx/templates/profile.j2
```diff
@@ -0,0 +1,20 @@
+server {
+    listen  80;
+    server_name {{ server.name }};
+
+    root /var/www/{{ server.name }}/web;
+
+    location / {
+        try_files $uri /app.php$is_args$args;
```
DRvanR: since this is dev only, this server config can be adjusted to point to `app_dev.php` exclusively. Makes our lives easier as then we don't have to include in the URLs, which coincidentally helps with SAML :wink:

DRvanR:
Some things need to be adjusted :smile:

ansible/development.yml
arothuis: Not sure if we need to save anything to DB (given the current knowns, I assume we don't); we can add it when we need to.

DRvanR: :+1:

rjkip:
After @DRvanR's suggestions, seems ok :+1:

arothuis:
@DRvanR @rjkip

DRvanR:
all seems to be in order now :+1:

rjkip:
:ship:

PR #3, opened 2015-10-26
Bugfix/php session permissions

arothuis:
Fixes bug where Exception is thrown (permission denied or path not found) when using (native) php sessions


arothuis:
@rjkip

rjkip:
:+1:

PR #4, opened 2015-10-26
Adds fontawesome (all formats) and Roboto (ttf)

arothuis:
Verified with BrowserStack:

* Roboto does not show up correctly in IE8 (however, the rest of Material Design Lite is also not displayed correctly), this is due to the lack of fallbacks to other font-types. 
* Roboto does show up correctly in IE9+
    * However, the rest of Material Design Lite seems to be displayed incorrectly due to flexbox.
* Font-awesome shows up correctly on both.


rjkip:
You could add Roboto support for IE8 by downloading [the EOT file](https://fonts.gstatic.com/s/roboto/v15/5YB-ifwqHP20Yn46l_BDhA.eot) and referencing it like Font-Awesome references their EOT file.

rjkip:
That seems like too much work. :+1:

PR #6, opened 2015-10-27
Removes last session_path usage from StateHandler class

arothuis:
Probably missed this because my session was already loaded.


arothuis:
@rjkip

rjkip:
:shipit:

PR #9, opened 2015-10-28
Removes default margin applied to footer-navigation

arothuis:


arothuis:
@rjkip, your expertise is required

rjkip:
:+1:

PR #10, opened 2015-10-29
Feature/create home page

arothuis:
[Ticket 106199686](https://www.pivotaltracker.com/story/show/106199686)


arothuis:
@rjkip :rotating_light:

app/Resources/views/base.html.twig
```diff
@@ -3,7 +3,7 @@
 <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>{% block title %}Profile{% endblock %}</title>
+    <title>{% block title %}OpenConext Profile{% endblock %}</title>
```
rjkip: Is the page title of every page OpenConext Profile? Should this actually be OpenConext rather than SURFconext, which is used throughout the translations?

rjkip: Also, I'd like to suggest to adjust the title to `{% block title %}{% endblock %} &ndash; {{ 'profile.app.name'|trans }}` and translate the app name to SURFconext. Then override the page titles in templates extending from `base.html.twig`.

src/OpenConext/ProfileBundle/Resources/views/Introduction/overview.html.twig
```diff
@@ -0,0 +1,16 @@
+{% extends 'OpenConextProfileBundle::layout.html.twig' %}
+
+{% block content %}
+    <h1>{{ 'introduction.long_title'|trans }}</h1>
```
rjkip: Translation keys are global identifiers, let's put some form of namespace before each key (`profile.`).

arothuis:
@rjkip, pretty please?

rjkip:
pretty

PR #11, opened 2015-10-30
Configure PHP-FPM to run on boot

arothuis:
[Bugfix, ticket 106924668](https://www.pivotaltracker.com/story/show/106924668)

Apparently, a startup script is not setup automatically by PHP-FPM. Chkconfig allows us to do so.
More info on [chkconfig](http://linux.die.net/man/8/chkconfig):
> chkconfig provides a simple command-line tool for maintaining the /etc/rc[0-6].d directory hierarchy by relieving system administrators of the task of directly manipulating the numerous symbolic links in those directories.

> (...)

>   If one of on, off, reset, or resetpriorities is specified after the service name, chkconfig changes the startup information for the specified service.

> (...)

PR #12, opened 2015-10-30
Uses service instead of command to start php-fpm on boot

arothuis:
[Ticket 106924668](https://www.pivotaltracker.com/story/show/106924668)

Improves #11 by using Ansible's (abstracted) service module instead of chkconfig.


arothuis:
@rjkip

rjkip:
:rocket:

PR #15, opened 2015-11-06
CSS and HTML fix

arothuis:
[Ticket 107178878](https://www.pivotaltracker.com/story/show/107178878)

Fixes:
* Roboto as font
* Margins in footer menu
* Image widths (giving the image a custom width might not be the best solution in the end, as it would probably be better to supply a correctly sized image).
* Spelling of 'SURFconext' (html)
* Removed hard-coded urls

These fixes will be taken into account when refactoring to [mdl components](https://www.pivotaltracker.com/story/show/107403918).


arothuis:
@rjkip

arothuis:
@DRvanR, @rjkip

app/config/global_view_parameters.yml
```diff
@@ -0,0 +1,10 @@
+parameters:
+    help_url:
+        en: "https://wiki.surfnet.nl/display/conextsupport/Profile+page"
+        nl: "https://wiki.surfnet.nl/display/conextsupport/Profile+page"
+    terms_of_service_url:
+        en: "https://wiki.surfnet.nl/display/conextsupport/Terms+of+Service+%%28EN%%29"
```
DRvanR: please document that `%` signs need to be escaped

src/OpenConext/ProfileBundle/Service/GlobalViewParameters.php
```diff
@@ -0,0 +1,116 @@
...
+                $translator->getLocale(),
+                explode(', ', $locales)
+            ));
+        }
+
+        $this->translator = $translator;
+        $this->locales = $locales;
+        $this->helpUrls = $helpUrls;
+        $this->platformUrls = $platformUrls;
+        $this->termsOfServiceUrls = $termsOfServiceUrls;
```
DRvanR: seem to be missing the check that a) for all locales a URL is defined and b) that all defined URLs belong to an existing locale

app/config/global_view_parameters.yml
```diff
@@ -0,0 +1,10 @@
+parameters:
+    help_url:
+        en: "https://wiki.surfnet.nl/display/conextsupport/Profile+page"
+        nl: "https://wiki.surfnet.nl/display/conextsupport/Profile+page"
```
rjkip: What about application name?

Could you try adding a `global_view_parameters.yml.dist` and letting incenteev/parameter-handler handle creation of the YAML file? This means other deploys of OpenConext-profile don't have to change files under version control. [Managing multiple ignored files](https://github.com/Incenteev/ParameterHandler/blob/f7b677a5f9bfd8bf5f165a0ff96ce7426b870312/README.md#managing-multiple-ignored-files)

app/Resources/translations/messages.en.xliff
```diff
@@ -11,10 +11,20 @@
...
       </trans-unit>
+      <trans-unit id="c01fb1e47571e92614557108db2ebb516dcdb7c0" resname="profile.application.platform_connection_description">
+        <jms:reference-file line="56">Resources/views/layout.html.twig</jms:reference-file>
+        <source>profile.application.platform_connection_description</source>
+        <target>This is a service connected through</target>
+      </trans-unit>
+      <trans-unit id="18e70a3e75cc5e9f61e4a379a265ae1bcdb71148" resname="profile.application.platform_connection_name">
+        <jms:reference-file line="56">Resources/views/layout.html.twig</jms:reference-file>
+        <source>profile.application.platform_connection_name</source>
+        <target>SURFConext</target>
```
rjkip: (SURFconext)

arothuis: D'oh

arothuis:
@rjkip, @DRvanR

src/OpenConext/ProfileBundle/Resources/config/services.yml
```diff
@@ -19,6 +19,17 @@ services:
         tags:
             - { name: twig.extension }
 
+    profile.global_view_parameters:
+        class: OpenConext\ProfileBundle\Service\GlobalViewParameters
+        factory: [OpenConext\ProfileBundle\Service\GlobalViewParameters, createForLocales]
```
DRvanR: why a factory?

src/OpenConext/ProfileBundle/Service/GlobalViewParameters.php
```diff
@@ -0,0 +1,159 @@
...
+    ) {
+        if (!in_array($translator->getLocale(), $locales)) {
+            throw new InvalidArgumentException(sprintf(
+                'Locale "%s" is not configured as a valid locale. Currently configured locales: %s',
+                $translator->getLocale(),
+                implode(', ', $locales)
+            ));
+        }
+
+        foreach ($locales as $locale) {
```
DRvanR: for future reference: `allKeyExists`

src/OpenConext/ProfileBundle/Service/GlobalViewParameters.php
```diff
@@ -0,0 +1,159 @@
...
+            Assert::keyExists($helpUrls, $locale, sprintf('Locale "%s" not configured for helpUrls', $locale));
+            Assert::keyExists($platformUrls, $locale, sprintf('Locale "%s" not configured for platformUrls', $locale));
+            Assert::keyExists($termsOfServiceUrls, $locale, sprintf(
+                'Locale "%s" not configured for termsOfServiceUrls',
+                $locale
+            ));
+        }
+
+        self::checkIfNoExtraLocales($helpUrls, $locales, 'helpUrls');
```
DRvanR: s/check/assert/g

src/OpenConext/ProfileBundle/Service/GlobalViewParameters.php
```diff
@@ -0,0 +1,159 @@
...
+        self::checkIfNoExtraLocales($platformUrls, $locales, 'platformUrls');
+        self::checkIfNoExtraLocales($termsOfServiceUrls, $locales, 'termsOfServiceUrls');
```
DRvanR: Thought Experiment: after verifying all locales we expect to be there are there, what would be the easiest way to determine whether or not there are _additional_ locales?

src/OpenConext/ProfileBundle/Service/GlobalViewParameters.php
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * @param array $configuration
+     * @param $locale
+     * @return string
+     */
+    private function resolveLocale(array $configuration, $locale)
+    {
+        if ($configuration[$locale]) {
+            return $configuration[$locale];
```
DRvanR: does this really warrant a helper method?

src/OpenConext/ProfileBundle/Service/GlobalViewParameters.php
arothuis: Checking their count would suffice. But that would be less explicit as it would depend on said additional test. In this case, that probably would not be a problem.

src/OpenConext/ProfileBundle/Service/GlobalViewParameters.php
arothuis: Ah, I could not find it in the docs, but you can prepend `all` to each assertion so an array of multiple values to be checked can be used.

src/OpenConext/ProfileBundle/Resources/config/services.yml
```diff
@@ -19,6 +19,17 @@ services:
...
+
+        arguments:
```
rjkip: Nit: remove double newline

src/OpenConext/ProfileBundle/Service/GlobalViewParameters.php
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * @var array
+     */
+    private $helpUrls;
+
+    /**
+     * @var array
+     */
+    private $platformUrls;
+    /**
```
rjkip: Nit: keep newlines between fields consistent. Either one or none.

src/OpenConext/ProfileBundle/Resources/config/services.yml
arothuis: Because the validation of the configuration parameters is not really part of the instantiation of the class that uses them. In other configurations, Symfony2's Configuration class handles this.

Also, we are using a `locales` variable that is not used inside the class, but only for validation. Giving it as a constructor argument would imply it is significant to the inner workings of the class.

src/OpenConext/ProfileBundle/Service/GlobalViewParameters.php
```diff
@@ -0,0 +1,159 @@
...
+    private $termsOfServiceUrls;
+
+    /**
+     * @param DataCollectorTranslator $translator
+     * @param array $helpUrls
+     * @param array $platformUrls
+     * @param array $termsOfServiceUrls
+     */
+    private function __construct(
+        DataCollectorTranslator $translator,
```
rjkip: Why this specific implementation? I guess this may actually fail in production, since it used for the web profiler.

arothuis:
@DRvanR, @rjkip

DRvanR:
fix the datasets, once that is done :+1:

rjkip:
:rocket::ship:

DRvanR:
:+1:

PR #22, opened 2015-11-26
Consent listing with filtered attributes

arothuis:
[Ticket 107811074](https://www.pivotaltracker.com/story/show/107811074)


arothuis:
@DRvanR @rjkip

src/OpenConext/Profile/Value/Consent.php
```diff
@@ -95,4 +113,12 @@ public function getConsentType()
     {
         return $this->consentType;
     }
+
+    /**
+     * @return null|AttributeSet
+     */
+    public function getAttributes()
+    {
+        return $this->attributes;
```
rjkip: Let's refrain from using nulls like all other places (ie. has-er). Also, maybe releasedAttributes or attributesReleased more clearly describes what the attributes mean in the context of a consent.

src/OpenConext/Profile/Value/ConsentList.php
```diff
@@ -35,11 +35,14 @@
     public function __construct(array $consents)
     {
         foreach ($consents as $consent) {
-            $this->initialiseWith($consent);
+            $this->initializeWith($consent);
```
rjkip: :(

src/OpenConext/Profile/Value/ConsentList.php
```diff
@@ -35,11 +35,14 @@
...
         }
     }
 
-    private function initialiseWith(Consent $consent)
+    /**
+     * @param Consent $consent
+     */
+    private function initializeWith(Consent $consent)
```
rjkip: I typically don't write a docblock if it doesn't have any added value. No need to remove it, but explaining my motives for not having written one :wink:

src/OpenConext/Profile/Value/ConsentType.php
```diff
@@ -26,7 +26,7 @@
     const TYPE_IMPLICIT = 'implicit';
 
     /**
-     * @var ConsentType
+     * @var string
```
rjkip: Good catch

src/OpenConext/ProfileBundle/Consent/BlacklistFilter.php
```diff
@@ -0,0 +1,62 @@
...
+use Surfnet\SamlBundle\SAML2\Attribute\Attribute;
+use Surfnet\SamlBundle\SAML2\Attribute\Filter\AttributeFilter as AttributeFilterInterface;
+
+final class BlacklistFilter implements AttributeFilterInterface
+{
+    /**
+     * @var array
+     */
+    const BLACKLIST = [
+        'urn:oid:2.5.4.42',
```
rjkip: You can be more specific by writing `@var string[]`. :wink:

Not sure if this qualifies as a constant, but since this list is unchanged since 2011, I think I'm ok with this. :stuck_out_tongue:

src/OpenConext/ProfileBundle/Consent/BlacklistFilter.php
```diff
@@ -0,0 +1,62 @@
...
+        'urn:oid:2.5.4.3',
+        'urn:oid:2.5.4.4',
+        'urn:oid:2.16.840.1.113730.3.1.241',
+        'urn:oid:0.9.2342.19200300.100.1.1',
+        'urn:oid:0.9.2342.19200300.100.1.3',
+        'urn:oid:1.3.6.1.4.1.1466.115.121.1.15',
+        'urn:oid:1.3.6.1.4.1.5923.1.1.1.6',
+        'coin:',
```
rjkip: Are these attribute prefixes, or could there be an attribute called `coin:`? In case of the former, it should be more clear that these are attribute prefixes, rather than exact matches. Perhaps `AttributePrefixBlacklistFilter` and `BLACKLISTED_ATTRIBUTE_PREFIXES` are more descriptive :no_mouth:

src/OpenConext/ProfileBundle/Consent/BlacklistFilter.php
arothuis: For now, it is constant. ;) We can make it configurable if necessary in the future.

src/OpenConext/ProfileBundle/Consent/BlacklistFilter.php
```diff
@@ -0,0 +1,62 @@
...
+     * @param Attribute $attribute
+     * @return bool
+     */
+    public function allows(Attribute $attribute)
+    {
+        $urnMace = $attribute->getAttributeDefinition()->getUrnMace();
+        $urnOid  = $attribute->getAttributeDefinition()->getUrnOid();
+
+        foreach (self::BLACKLIST as $blacklistedAttribute) {
+            if (strpos($urnMace, $blacklistedAttribute) !== false || strpos($urnOid, $blacklistedAttribute) !== false) {
```
rjkip: Perhaps a `strpos === 0` is more correct if we're dealing with prefixes here, rather than "could-be-anywhere-in-the-string" things.

src/OpenConext/ProfileBundle/Resources/config/services.yml
```diff
@@ -81,3 +81,19 @@ services:
...
+
+    profile.service.consent_listing:
+        class: OpenConext\ProfileBundle\Service\ConsentListingService
+        arguments:
+            - @profile.user.user_provider
+            - @profile.service.consent
+            - @profile.consent.blacklist_filter
+
+
+    profile.consent.blacklist_filter:
```
rjkip: Nit: double newline

src/OpenConext/ProfileBundle/Resources/views/MyServices/overview.html.twig
```diff
@@ -67,7 +67,7 @@
                                 </tr>
                                 </thead>
                                 <tbody>
-                                {% for attribute in app.user.attributes %}
+                                {% for attribute in consent.attributes %}
```
rjkip: If `getAttributes()` had a has-er and would throw when attributes=null, we'd get a LogicException here, which'd be good.

src/OpenConext/Profile/Value/Consent.php
arothuis: Released attributes sounds fine to me, as `releasing` is what is consented to. Has-er + LogicException seems the way to go with a getter like this.

src/OpenConext/ProfileBundle/Service/ConsentListingService.php
```diff
@@ -0,0 +1,83 @@
...
+        $user               = $this->userProvider->getCurrentUser();
+        $consentList        = $this->consentService->findAllFor($user);
+        $filteredAttributes = $user->getAttributes()->apply($this->attributeFilter);
+
+        $newConsents = [];
+
+        /** @var Consent $consent */
+        foreach ($consentList as $consent) {
+            $newConsent = $consent->givenFor($filteredAttributes);
+            $newConsents[] = $newConsent;
```
rjkip: `array_map` ftw!

rjkip: Ah sorry, you start off with a ConsentList. Maybe a `ConsentList::map(callable $predicate):ConsentList` would be nice?

src/OpenConext/Profile/Value/Consent.php
DRvanR: Disagree, it is a value object. Therefor, it has a known value, so either it should be split out to a new attribute  (eeehr, something like ConsentWithAttributes), or the consumer should be responsible for ensuring compatibility with the known api: `null|AttributeSet`. Has-er may be added for convenience, but certainly not a LogicException - it is documented that a Consent can exist with or without Attrributes, so assuming attributes exists at all times conflicts with documented behavior.

A different object has my preference btw :tongue:

src/OpenConext/ProfileBundle/Consent/BlacklistFilter.php
DRvanR: This is the BC version, hence this should be kept.

src/OpenConext/ProfileBundle/Consent/BlacklistFilter.php
DRvanR: It are attributes names or part of attribute names :disappointed:

src/OpenConext/ProfileBundle/Resources/views/MyServices/overview.html.twig
DRvanR: (hence my preference for a new different object for consent with attributes, so we known it always returns an iterable)

src/OpenConext/ProfileBundle/Service/ConsentListingService.php
DRvanR: wouldn't do that here (now) to be honest, as it can become more complex when ARP is included which is fetched/enforced per SP/user combo (so per consent). Kinda scopingissue in when calling API's in a predicate in this setup.

src/OpenConext/Profile/Value/Consent.php
rjkip: The above suggestion wasn't suggested in the assumption that the released attributes are present at all times. That's what the setter is for.

I'd also like a different object, but I assumed the `givenFor()` approach was as discussed with you @DRvanR. My bad.

Nothing wrong with "released attributes" vs. "attributes", though? :smiley:

src/OpenConext/ProfileBundle/Consent/BlacklistFilter.php
rjkip: Very well.

src/OpenConext/ProfileBundle/Resources/views/MyServices/overview.html.twig
rjkip: :+1:

src/OpenConext/ProfileBundle/Service/ConsentListingService.php
rjkip: Very well.

DRvanR: FYI: we just discussed a different implementation (the separate object) and due to splitting out responsibilities, the `:map` call will actually be added :wink:

src/OpenConext/Profile/Value/Consent.php
DRvanR: oh, it was discussed, just evolving design and me making a general remark :wink:

arothuis:
@DRvanR @rjkip :shower:

src/OpenConext/ProfileBundle/Controller/MyServicesController.php
```diff
@@ -31,38 +33,42 @@ class MyServicesController
...
+        $this->userProvider                = $userProvider;
+        $this->specifiedConsentListService = $specifiedConsentListService;
     }
 
     public function overviewAction()
     {
-        $user        = $this->tokenStorage->getToken()->getUser();
-        $consentList = $this->consentService->findAllFor($user);
+        if (!$this->userProvider->hasCurrentUser()) {
+            throw new RuntimeException('Cannot retrieve MyServices overview: no user available');
```
DRvanR: This should not be done, rather a guard should be added that this method may only be invoked when there is a user logged in, which we enforce through the security config and verify here (see [the relevant issue][1]). I'd rather put in a comment with a todo than this tbh.

[1]: https://www.pivotaltracker.com/story/show/107020530

DRvanR:
one last thing to discuss, and a nit to squash other than that :+1:

src/OpenConext/ProfileBundle/Controller/MyServicesController.php
arothuis: It felt weird when I put it there. Indeed, the whole application (except of course logging in/saml and the public metadata) assumes someone to be logged in so it should be enforced at that level, but I felt some defense was needed in any case (so misconfiguration or misuse of components will be easily detected). 

Will make it a todo for now so we can pick it up later.

rjkip:
Let's-a go :supermario:

PR #23, opened 2015-11-27
Removes unused session in StateHandler

arothuis:
When removing the session migration from the Saml StateHandler, the session object itself has not been removed. This PR cleans that up.


arothuis:
@rjkip, @DRvanR

rjkip:
How could we forget :smile: :+1:

DRvanR:
yes please! :+1:

PR #26, opened 2015-11-30
Secure application

arothuis:
[Ticket 107020530](https://www.pivotaltracker.com/story/show/107020530)


arothuis:
@rjkip, @DRvanR

app/config/config.yml
```diff
@@ -26,6 +26,44 @@ framework:
...
+            '^/.*': DENY
+    external_redirects:
+        abort: true
+        log: true
+        whitelist:
+            - %saml_remote_idp_host%
+
+    csp:
+        report:
+            report-uri: [/nelmio/csp/report]
```
rjkip: Do you have a controller set up for these CSP reports (see links below)? Can there be multiple report URLs? Also, do we only enforce `script-src: [self]` (JavaScript)? What about `default-src: [self]` (everything)?

https://github.com/SURFnet/Stepup-SelfService/blob/d415fc9e81d1b0663a85f867b10ce8461345aee3/app/config/config.yml#L71
https://github.com/SURFnet/Stepup-SelfService/blob/1ba8b776aff73aaa490058420de06d5e105c0206/app/config/routing.yml#L6

src/OpenConext/ProfileBundle/Security/Guard.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public function __construct(AuthorizationCheckerInterface $authorizationChecker)
+    {
+        $this->authorizationChecker = $authorizationChecker;
+    }
+
+    public function userIsLoggedIn()
+    {
+        if (!$this->authorizationChecker->isGranted('ROLE_USER')) {
+            throw new AccessDeniedHttpException('User should be logged in to use this service');
```
rjkip: I think we should either throw a [AccessDeniedException](https://github.com/symfony/symfony/blob/0055887cbdf89d9b3b0235f2089a2751bcb52d17/src/Symfony/Component/Security/Core/Exception/AccessDeniedException.php), which Symfony [converts](https://github.com/symfony/symfony/blob/0055887cbdf89d9b3b0235f2089a2751bcb52d17/src/Symfony/Component/Security/Http/Firewall/ExceptionListener.php#L119) to an AccessDeniedHttpException, or move the Guard to the namespace `...\Security\Http`.

app/config/config.yml
DRvanR: `none` for object and `[self]` for the rest would be better indeed :stuck_out_tongue_winking_eye: If we need something different in dev, we can configure that there.

And good point about the CSP enpoint, the report uri currently used is the default nelmio provided one. Would like to have the linked logging version though.

src/OpenConext/ProfileBundle/Security/Guard.php
DRvanR: The former indeed, `AccessDeniedException`.

arothuis:
@rjkip, @DRvanR

rjkip:
:+1:

DRvanR:
nice nice :+1:

PR #29, opened 2015-12-09
Document global view parameters

arothuis:
These need to be added to OpenConext-deploy as well, but first they need to be documented a bit.


arothuis:
@rjkip @DRvanR

DRvanR:
:+1:

rjkip:
:+1:

PR #30, opened 2015-12-09
Strips mailto scheme from email address

arothuis:
[Ticket 106478068](https://www.pivotaltracker.com/story/show/106478068)


arothuis:
This was the wrong approach. Will create a different PR.

PR #45, opened 2016-06-02
Add and document makeRelease.sh script and composer command

arothuis:
Added simple script, based on existing scripts within the platform, to create deployment tarballs.
Added documentation.

### DIFF
--- a/makeRelease.sh
+++ b/makeRelease.sh
@@ -54,13 +54,9 @@ echo "Commit: ${COMMITHASH}" >> ${PROJECT_DIR}/RELEASE &&
 echo "Cleaning build of dev files" &&
 rm -rf ${PROJECT_DIR}/ansible &&
 rm -rf ${PROJECT_DIR}/Vagrantfile &&
-rm -rf ${PROJECT_DIR}/.idea &&
 rm -rf ${PROJECT_DIR}/.git &&
 rm -f ${PROJECT_DIR}/.gitignore &&
 rm -f ${PROJECT_DIR}/makeRelease.sh &&
-rm -f ${PROJECT_DIR}/bin/composer.phar &&
-rm -rf ${PROJECT_DIR}/features &&
-rm -rf ${PROJECT_DIR}/behat.yml &&
 rm -rf ${PROJECT_DIR}/.rmt.yml &&
 rm -rf ${PROJECT_DIR}/RMT &&
 rm -rf ${PROJECT_DIR}/.scrutinizer.yml &&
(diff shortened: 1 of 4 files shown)


thijskh:
This looks good to me!

makeRelease.sh
```diff
@@ -0,0 +1,106 @@
...
+
+echo "Tagging the release in RELEASE file" &&
+COMMITHASH=`git rev-parse HEAD` &&
+echo "Tag: ${TAG}" > ${PROJECT_DIR}/RELEASE &&
+echo "Commit: ${COMMITHASH}" >> ${PROJECT_DIR}/RELEASE &&
+
+echo "Cleaning build of dev files" &&
+rm -rf ${PROJECT_DIR}/ansible &&
+rm -rf ${PROJECT_DIR}/Vagrantfile &&
+rm -rf ${PROJECT_DIR}/.idea &&
```
DRvanR: is this in the project when doing a checkout? Should be `.gitignore`-ed :wink:

makeRelease.sh
```diff
@@ -0,0 +1,106 @@
...
+rm -rf ${PROJECT_DIR}/.git &&
+rm -f ${PROJECT_DIR}/.gitignore &&
+rm -f ${PROJECT_DIR}/makeRelease.sh &&
+rm -f ${PROJECT_DIR}/bin/composer.phar &&
```
DRvanR: don't think we have this, or the next 2 files? (`features` and `behat.yml`)

makeRelease.sh
```diff
@@ -0,0 +1,106 @@
...
+rm -rf ${PROJECT_DIR}/.travis.yml &&
+rm -rf ${PROJECT_DIR}/.travis.php.ini &&
+rm -f ${PROJECT_DIR}/web/app_dev.php &&
+
+echo "Removing application cache, logs, bootstrap and parameters" &&
+rm -f ${PROJECT_DIR}/app/bootstrap.php.cache &&
+rm -rf ${PROJECT_DIR}/app/cache/* &&
+rm -rf ${PROJECT_DIR}/app/logs/* &&
+rm -f ${PROJECT_DIR}/app/config/global_view_parameters.yml &&
+rm -f ${PROJECT_DIR}/app/config/parameters.yml &&
```
DRvanR: these two parameter files only have `.dist` variants right?

DRvanR:
a few nits, other than that :+1:

makeRelease.sh
arothuis: I copied this from another `makeRelease`. I know there are people that don't have it in their global gitignore, so I left it in, but I agree: they should be gitignored.

makeRelease.sh
arothuis: Nope. Should've removed it.

makeRelease.sh
arothuis: Ah right, they are .gitignored so they should be removed.

Issue #46, opened 2016-06-06
Trouble dealing with multivalued attributes


thijskh:
Profile has a problem dealing with multivalued attributes. Logging into profile with a common set of attributes therefore fails with:
```
request.CRITICAL: Uncaught PHP Exception UnexpectedValueException: "AttributeDefinition "eduPersonEntitlement" has a single-value multiplicity, got "2" values" at /opt/openconext/OpenConext-profile-0.2.0/vendor/surfnet/stepup-saml-bundle/src/SAML2/Attribute/Attribute.php line 44
```

Most likely, the best way to fix this, is to upgrade to stepup-saml-bundle >= 2 which removed the concept of attribute multiplicity. Profile seems to work with this, except for attributes like this:

```
            <saml:Attribute Name="urn:mace:dir:attribute-def:eduPersonTargetedID"
                            NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:uri"
                            >
                <saml:AttributeValue>
                    <saml:NameID Format="urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified">urn:collab:person:surfnet.nl:kinkhorst</saml:NameID>
                </saml:AttributeValue>
            </saml:Attribute>
```
which result in doubly nested arrays which the twig templates cannot deal with currently.

arothuis:
It does look like an error in how multiplicity is dealt with and `simple-saml-bundle >= 2` indeed fixes this.
The template could then be rendered by doing a nested for-loop.

However, I see an empty `DOMNodeList` as a value for the `eduPersonTargetedID` in my reproduction of the issue and it will therefore show up blank when the template-solution would be implemented.

I find this strange for two reasons:

1. A `DOMNodeList` is IMHO not really a useful representation of an attribute. Other attributes are scalar/strings, while a `DOMNodeList` is specific for how XML parsing is handled internally.
2. The `DOMNodeList` should contain a NameID in my reproduction of the issue, because I see a NameID in the SAML request.

At first sight, this is due to how `simple-saml-bundle` deals with this kind of nesting. An adequate solution would require some more research.

thijskh:
Thanks. For profile, I would be fine with just not displaying these kinds of attributes at all (for now), as they are generated by OpenConext. If we upgrade the lib and skip any attribute that is not a simple type, we have a useful and functioning profile.

The issue in `stepup-saml-bundle` is indeed something to discuss there, perhaps turn that into an Issue?

[...]

PR #49, opened 2016-06-09
Bugfix: parsing attributes

arothuis:
See: #46

### DIFF
--- a/src/OpenConext/Profile/Entity/AuthenticatedUser.php
+++ b/src/OpenConext/Profile/Entity/AuthenticatedUser.php
@@ -20,6 +20,8 @@ namespace OpenConext\Profile\Entity;
 
 use OpenConext\Profile\Assert;
 use OpenConext\Profile\Value\EntityId;
+use RuntimeException;
+use Surfnet\SamlBundle\SAML2\Attribute\Attribute;
 use Surfnet\SamlBundle\SAML2\Attribute\AttributeSet;
 use Surfnet\SamlBundle\SAML2\Response\AssertionAdapter;
 
@@ -43,13 +45,47 @@ final class AuthenticatedUser
     /**
      * @param AssertionAdapter $assertionAdapter
      * @param EntityId[] $authenticatingAuthorities
+     *
      * @return AuthenticatedUser
+     * @throws RuntimeException
      */
     public static function createFrom(AssertionAdapter $assertionAdapter, array $authenticatingAuthorities)
     {
+        $attributes = [];
+
+        /** @var Attribute $attribute */
+        foreach ($assertionAdapter->getAttributeSet() as $attribute) {
+            $definition = $attribute->getAttributeDefinition();
+
+            // We need to replace the eduPersonTargetedID attribute value as that is a nested NameID attribute.
+            if ($definition->getName() === 'eduPersonTargetedID') {
+
+                /** @var \DOMNodeList[] $eptiValues */
+                $eptiValues = $attribute->getValue();
+                $eptiDomNodeList = $eptiValues[0];
+
+                if (!$eptiDomNodeList instanceof \DOMNodeList || $eptiDomNodeList->length !== 1) {
+                    throw new RuntimeException(
+                        sprintf(
+                            'EPTI attribute must contain exactly one NameID element as value, received: %s',
+                            print_r($eptiValues, true)
+                        )
+                    );
+                }
+
+                $eptiValue  = $eptiDomNodeList->item(0);
+                $eptiNameId = \SAML2_Utils::parseNameId($eptiValue);
+
+                $attributes[] = new Attribute($definition, [$eptiNameId['Value']]);
+                continue;
+            }
+
+            $attributes[] = $attribute;
+        }
+
         return new self(
             $assertionAdapter->getNameId(),
-            $assertionAdapter->getAttributeSet(),
+            AttributeSet::create($attributes),
             $authenticatingAuthorities
         );
     }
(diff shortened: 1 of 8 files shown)


src/OpenConext/Profile/Entity/AuthenticatedUser.php
```diff
@@ -43,13 +45,47 @@
...
+
+            // We need to replace the eduPersonTargetedID attribute value as that is a nested NameID attribute.
+            if ($definition->getName() === 'eduPersonTargetedID') {
+
+                /** @var \DOMNodeList[] $eptiValues */
+                $eptiValues = $attribute->getValue();
+                $eptiDomNodeList = $eptiValues[0];
+
+                if (!$eptiDomNodeList instanceof \DOMNodeList || $eptiDomNodeList->length !== 1) {
+                    throw new RuntimeException(
```
DRvanR: Please use a specific exception (even an `\OpenConext\Profile\RuntimeException` would be better) so that when handling exceptions it is easier to differentiate between Profile exceptions and other exceptions.

DRvanR:
one nit, other than that :+1:

src/OpenConext/Profile/Entity/AuthenticatedUser.php
```diff
@@ -43,13 +45,47 @@
...
+                }
+
+                $eptiValue  = $eptiDomNodeList->item(0);
+                $eptiNameId = \SAML2_Utils::parseNameId($eptiValue);
+
+                $attributes[] = new Attribute($definition, [$eptiNameId['Value']]);
+                continue;
+            }
+
+            $attributes[] = $attribute;
```
rjkip: I'd suggest inverting the if statement, so that, when reading this code, you can quickly take the easy path `$attributes[]=$attribute;` out this for loop.

src/OpenConext/Profile/Entity/AuthenticatedUser.php
```diff
@@ -43,13 +45,47 @@
...
+
+                /** @var \DOMNodeList[] $eptiValues */
+                $eptiValues = $attribute->getValue();
+                $eptiDomNodeList = $eptiValues[0];
+
+                if (!$eptiDomNodeList instanceof \DOMNodeList || $eptiDomNodeList->length !== 1) {
+                    throw new RuntimeException(
+                        sprintf(
+                            'EPTI attribute must contain exactly one NameID element as value, received: %s',
+                            print_r($eptiValues, true)
```
rjkip: Did you verify that print_r'ing an array of DOMNodeLists yields useful information? Also: I don't feel like digging up the context for this, but could NameIDs include PII (name and institution) which should not end up in the logs? (Exception messages are likely to be logged.)

arothuis: Nice find! This approach is comparable to [Stepup-Gateway's][gateway], although I'm not sure if PII would be present there. I guess it depends on what type of NameIDs are to be expected.

Regarding the `print_r`:
Whether or not an array of DOMNodeLists is returned is not certain, so what *is* returned would be interesting to note. Furthermore, when a `DOMNodeList` is returned, its length property should be exposed (which is something interesting as well).

We could, instead, note the type of `$eptiDomNodeList` and -- if it is a DomNodeList -- its length. That might be a bit complex for a simple Exception, but would prevent PII (if any) to be leaked.

[gateway]: https://github.com/SURFnet/Stepup-Gateway/blob/383ca16200378de38bac098717ba9ed7a3999e5e/src/Surfnet/StepupGateway/GatewayBundle/Service/ProxyResponseService.php#L203-L207

rjkip: > This approach is comparable to Stepup-Gateway's, although I'm not sure if PII would be present there.

The Gateway copies the `eduPersonTargetedId` attribute to the assertion's NameID. So I would derive that this is the NameID that does not contain for example a name. @DRvanR can you verify? In any case, perhaps it would just be safer not to log attribute values in general.

> We could, instead, note the type of $eptiDomNodeList and -- if it is a DomNodeList -- its length. That might be a bit complex for a simple Exception, but would prevent PII (if any) to be leaked.

Yes. That would be two if statements with different messages.

Should we maybe test this quite complex method? (Sorry.)

DRvanR: The gateway is SSA context, this is OpenConext. Logging PII already can happen in an OpenConext platform (e.g. in EngineBlock), not sure if that in general would be an issue. I do agree that it should be avoided though. What the NameID does or does not contain should not be a matter as it is an identifier that potentially contains PII which we cannot control.

As to how to log this, make a custom exception with a factory method that solves that, as creating a specific exception message for a specific value arguably is not really a concern that should be tackled here.

arothuis:
Thoughts after last updates, @rjkip, @DRvanR ?

DRvanR:
:+1:

src/OpenConext/Profile/Entity/AuthenticatedUser.php
pmeulen: Very good to see that you take PII into consideration! Note that in case of error the fingers cross logging is expected to log the SAML messages (which in this case contain the same PII). 
Generally speaking: 
- log retention is limited in time (for use 6 max months). Logs are one of the easiest locations to implement 
"automatically forget".
- Fingers crossed logging works a limiter bacause it only logs additional info in case of error
- The trade-off is between us being able to support/troubelshoot issues and not wanting to limit PII.

rjkip:
Awesome :ok_hand:

Issue #47, opened 2016-06-09
Cookie parameters language cookie not correct


thijskh:
Profile sets a `lang` cookie which is a session cookie with the secure and httponly flags on. This conflicts with the global `lang` cookie shared between other OpenConext applications which is valid for now+2 months; and does not have the secure or httponly flags set.

Please change profile's cookie settings to be in line with this (for the `lang` cookie).

thijskh:
![cookie settings; the first one is 'correct'](https://cloud.githubusercontent.com/assets/3808792/15925539/dbb4d380-2e37-11e6-88f0-a3c57d68cf7e.png)

arothuis:
This has been made configurable in #50

PR #50, opened 2016-06-10
OpenConext compatible locale cookie storage

arothuis:
See: #47

### DIFF
--- a/src/OpenConext/ProfileBundle/DependencyInjection/OpenConextProfileExtension.php
+++ b/src/OpenConext/ProfileBundle/DependencyInjection/OpenConextProfileExtension.php
@@ -23,6 +23,7 @@ use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Component\DependencyInjection\Loader;
+use \DateTime;
 
 class OpenConextProfileExtension extends Extension
 {
@@ -44,7 +45,7 @@ class OpenConextProfileExtension extends Extension
         $this->parseLocaleCookieStorageConfiguration(
             $config['locale_cookie_domain'],
             $config['locale_cookie_key'],
-            $container->getDefinition('profile.locale.cookie_expiration_date'),
+            $config['locale_cookie_expires_in'],
             $config['locale_cookie_secure'],
             $config['locale_cookie_http_only'],
             $container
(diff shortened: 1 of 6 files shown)


src/OpenConext/ProfileBundle/DependencyInjection/OpenConextProfileExtension.php
```diff
@@ -41,9 +41,12 @@ public function load(array $configs, ContainerBuilder $container)
 
         $this->parseDefaultLocaleConfiguration($config['default_locale'], $container);
         $this->parseAvailableLocaleConfiguration($config['locales'], $container);
-        $this->parseCookieStorageConfiguration(
+        $this->parseLocaleCookieStorageConfiguration(
```
rjkip: Nice.

src/OpenConext/ProfileBundle/Resources/config/services.yml
```diff
@@ -1,4 +1,10 @@
 services:
+    profile.locale.cookie_expiration_date:
+        public: false
+        class: DateTime
+        calls:
+            - [modify, ['+2 days']]
```
rjkip: The original issue states 2 months.

src/OpenConext/ProfileBundle/Storage/SingleCookieStorage.php
```diff
@@ -42,13 +43,43 @@ class SingleCookieStorage implements EventSubscriberInterface
...
+
+    /**
+     * @var boolean
+     */
+    private $cookieHttpOnly;
+
+    public function __construct(
+        $cookieDomain,
+        $cookieKey,
+        DateTime $cookieExpirationDate = null,
```
rjkip: You can use `DateTimeInterface` here.

While writing this comment, I found that the only place where the supported PHP versions are "documented" is in the Travis build file. Profile should require specific PHP versions in its Composer file to prevent installing incompatible packages (for example PHP 7-only packages). The README should also state that only PHP 5.6 is officially supported. PHP 7 is only tested for forward compatibility. Would you create an issue for that? :wink:

src/OpenConext/ProfileBundle/Resources/config/services.yml
arothuis: LOL. Misread that, thanks :)

src/OpenConext/ProfileBundle/Storage/SingleCookieStorage.php
```diff
@@ -42,13 +43,43 @@ class SingleCookieStorage implements EventSubscriberInterface
...
+        $cookieSecure = true,
+        $cookieHttpOnly = true
```
DRvanR: These defaults do not match the defaults of the configuration, perhaps make them the same?

src/OpenConext/ProfileBundle/Storage/SingleCookieStorage.php
```diff
@@ -42,13 +43,43 @@ class SingleCookieStorage implements EventSubscriberInterface
...
+    ) {
         Assert::string($cookieDomain);
         Assert::string($cookieKey);
+        Assert::boolean($cookieSecure);
+        Assert::boolean($cookieHttpOnly);
```
DRvanR: would add some specific messages "Expected value to be a boolean, NULL given" for instance is not really helpful when there are multiple boolean checks.

src/OpenConext/ProfileBundle/Storage/SingleCookieStorage.php
arothuis: Defaults of the configuration are specific to storing locales, this is a general cookie storage. I chose to make secure cookies the default, but I could use the Cookie object's defaults.

arothuis:
Thoughts after last updates, @rjkip, @DRvanR ?

DRvanR:
:+1:

rjkip:
:ok_hand:

PR #57, opened 2016-09-23
Turn on security checker and do composer update

relaxnow:
At Ibuildings we would like to take a more proactive role at monitoring and alerting for security updates of the software we create and deliver to customers.

As part of this, for this project we'd like to:

* Upgrade to the latest patch level of Symfony 2.7
* Turn on the security checker and make it fail pull requests.

Unfortunately I had to do a full composer update as Composer got lost with all the interdependencies.
Please see the commit for a full overview of all upgraded packages.

Also we'd like to turn on nightly builds ( https://docs.travis-ci.com/user/cron-jobs/ ), though I think I can request this from Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,8 @@ cache:
 before_script:
   - phpenv config-add .travis.php.ini
   - composer self-update
-  - composer install --prefer-dist
+  - if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.6" ]]; then composer install --prefer-dist -n -o; fi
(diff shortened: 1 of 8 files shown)


relaxnow:
Yes, thanks @thijskh, I'll look into this probably next friday.

arothuis:
Constraining the PHP version in composer to PHP 5.6, always makes Travis fail on PHP 7.
We could configure Travis to run composer with `--ignore-platform-reqs` when running as PHP 7, but without that flag when it is running PHP 5.6.

PR #60, opened 2016-09-26
Update profile interface texts

thijskh:
As provided by @arnoutt.


arothuis:
Looking good. Merging.

Issue #54, opened 2016-08-16
Do not set HSTS header


thijskh:
We determine globally in OpenConext-deploy which vhosts get HSTS-headers and with what settings. It would help if profile would not send a header of its own (so we don't have to remove it).

This probably boils down to removing from `app/config/config.yml`:
```yaml
nelmio_security:
    …
    forced_ssl:
        hsts_max_age: 2592000 # 30 days
        hsts_subdomains: true
```

arothuis:
Fixed in #64

PR #64, opened 2016-10-14
Drop setting HSTS header

arothuis:
This should be configured on the server level in production, see OpenConext-Deploy

#54

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -50,6 +50,8 @@ nelmio_security:
     xss_protection:
         enabled: true
         mode_block: true
+    forced_ssl:
+        enabled: true
 
 monolog:
     handlers:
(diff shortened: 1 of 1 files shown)


thijskh:
:+1:

thijskh:
:+1:

thijskh:
As long as this doesn't break running Profile behind a load balancer where Profile itself is accessed over http from the frontend.

arothuis:
According to the [NelmioSecurityBundle documentation](https://github.com/nelmio/NelmioSecurityBundle#forced-httpsssl-handling), it will "redirect all users reaching the site with a http:// URL to a https:// URL". So if you cannot deal with `https://profile.*`, this will not work.

I think you could also enforce "force_ssl" using webserver configuration, for instance by using rewrite rules. If we do not want to require the application-level "enforcement" of SSL, we could drop this configuration.

thijskh:
Enabling it doesn't seem to cause trouble in our test environment. So :+1: from me.

arothuis:
Thanks for testing it!

Issue #58, opened 2016-09-26
Add some links to texts


thijskh:
@arnoutt writes:
- Onder My Profile heb ik een link in de tekst opgenomen, graag omzetten naar link
- Onder My Profile staat dat de gebruiker z'n instelling moet mailen, hier moet dan ook het support contact van z'n IdP worden getoond
- Onder My Profile staat 'please refer to the 'My Services' tab.'. Daar ook graag linken naar My Services.

arothuis:
I'd rather not allow raw HTML in translations as one would get too much control over HTML and JavaScript (with the potential for XSS). One solution would be to offer custom tags like old-school forums used to have (bbcode), i.e: `[url]http://www.example.com[/url]` and `[url="http://www.example.com"]A link[/url]`. We could implement this in a our own Twig function and make it extensible by abstracting the replacement strategies. Alternatively, we could look for and use a `bbcode` or `markdown` (or whatever) Twig extension.

Thoughts, @DRvanR?

arothuis:
Discussed in IRL: We should use a Twig sandbox in which certain HTML attributes are whitelisted.

arnoutt:
Sounds good to me!

arothuis:
@arnoutt, what do you mean by "Onder My Profile staat dat de gebruiker z'n instelling moet mailen, hier moet dan ook het support contact van z'n IdP worden getoond".

At the moment, the users' support contact information [is displayed already] (https://github.com/OpenConext/OpenConext-profile/blob/develop/src/OpenConext/ProfileBundle/Resources/views/MyProfile/overview.html.twig#L13-L16). Should something be changed here?

arothuis:
Bar previous question, this has been implemented in #66.

arnoutt:
I see, it looks like the entire sentence is not shown when the contact data is not known.

I prefer though to always show the entire sentence, and if no contact information is known change the text to:

[EN] If you have any questions about your personal data, please contact the help desk of your institution.
[NL] Heb je vragen over je persoonsgegevens? Neem dan contact op met de helpdesk van je instelling.

arothuis:
Thanks. Will add the message for when a support contact is missing.

arothuis:
Implemented in #66 and #67.

PR #66, opened 2016-10-18
Enrich Profile texts and translations with more links

arothuis:
This implements #58.

### DIFF
--- a/app/Resources/translations/messages.en.xliff
+++ b/app/Resources/translations/messages.en.xliff
@@ -1,321 +1,326 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
-  <file date="2016-09-23T16:30:08Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
+  <file date="2016-10-18T16:02:41Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
     <header>
       <tool tool-id="JMSTranslationBundle" tool-name="JMSTranslationBundle" tool-version="1.1.0-DEV"/>
       <note>The source node in most cases contains the sample message as written by the developer. If it looks like a dot-delimitted string such as "form.label.firstname", then the developer has not provided a default message.</note>
     </header>
     <body>
       <trans-unit id="4422839d75319cb038791289059a5c37b6464d3e" resname="profile.application.header">
-        <jms:reference-file line="20">Resources/views/layout.html.twig</jms:reference-file>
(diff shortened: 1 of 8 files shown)


arothuis:
This is the simplest way of safely adding links to the texts. If we need a more flexible approach, we should consider implementing some sort of HTML whitelist.

arothuis:
@rjkip I added `stiptags` because using `raw` instead would make `replace` pointless as you would just accept HTML at that point.

arothuis:
Removed striptags in favor of a more escape-all approach.

Would you mind reviewing this, @rjkip?

app/Resources/translations/messages.en.xliff
```diff
@@ -1,321 +1,326 @@
...
     <header>
       <tool tool-id="JMSTranslationBundle" tool-name="JMSTranslationBundle" tool-version="1.1.0-DEV"/>
       <note>The source node in most cases contains the sample message as written by the developer. If it looks like a dot-delimitted string such as "form.label.firstname", then the developer has not provided a default message.</note>
     </header>
     <body>
       <trans-unit id="4422839d75319cb038791289059a5c37b6464d3e" resname="profile.application.header">
-        <jms:reference-file line="20">Resources/views/layout.html.twig</jms:reference-file>
         <source>profile.application.header</source>
         <target>Overview of your SURFconext profile</target>
+        <jms:reference-file line="20">/../src/OpenConext/ProfileBundle/Resources/views/layout.html.twig</jms:reference-file>
```
rjkip: What the heck is happening here? :sweat_smile:

src/OpenConext/ProfileBundle/Resources/views/MyProfile/overview.html.twig
```diff
@@ -6,8 +6,8 @@
 
 {% block content %}
     <h1>{{ 'profile.my_profile.long_title'|trans }}</h1>
-    <p>{{ 'profile.my_profile.introduction'|trans }}</p>
-    <p>{{ 'profile.my_profile.explanation'|trans }}</p>
+    <p>{{ 'profile.my_profile.introduction'|trans }} <a href="{{ global_view_parameters.attributeInformationUrl }}">{{ 'profile.my_profile.attributes_information_link_title'|trans }}</a>.</p>
+    <p>{{ 'profile.my_profile.explanation'|trans|escape|replace({'%my_services_link%': '<a href="' ~ path('profile.my_services_overview') ~ '">' ~ 'profile.my_services.short_title'|trans ~ '</a>'})|raw }}</p>
```
rjkip:
```diff
- ~ 'profile.my_services.short_title'|trans ~
+ ~ 'profile.my_services.short_title'|trans|e ~
```

Issue #55, opened 2016-08-23
Should not throw error on unknown attribute


thijskh:
When a user has an attribute that is unknown to profile (or: to stepup-saml-bundle), it throws a 500 error into the user's face.

```
Aug 22 19:15:09 papp3 openconext-profile[25780]: request.CRITICAL: Uncaught PHP Exception Surfnet\SamlBundle\Exception\UnknownUrnException: "Urn "urn:schac:attribute-def:schacPersonalUniqueCode" has not been defined in the AttributeDictionary" at /opt/openconext/OpenConext-profile-0.3.0/vendor/surfnet/stepup-saml-bundle/src/SAML2/Attribute/AttributeDictionary.php line 130 {"exception":"[object] (Surfnet\\SamlBundle\\Exception\\UnknownUrnException(code: 0): Urn \"urn:schac:attribute-def:schacPersonalUniqueCode\" has not been defined in the AttributeDictionary at /opt/openconext/OpenConext-profile-0.3.0/vendor/surfnet/stepup-saml-bundle/src/SAML2/Attribute/AttributeDictionary.php:130)"} []
```

Just adding this specific attribute to the AttributeDictionary is not sufficient. Since we pass all attributes we get from the IdP on to profile, there can be any attribute that happens to be unknown to us in the set and it's not friendly to error on that. Instead, the attribute should probably just be presented in the list, but instead of with a "nice" name, with its actual name.

arothuis:
A solution in the Stepup-saml-bundle will offer Profile a way to allow fallbacks in the created AttributeSet. 

An attribute consists of a definition (a name, urn:mace and urn:oid) and a value. How should we deal with fallback attributes? We could create a definition during runtime for the given attribute: setting the same name, urn:mace and urn:oid as the received attribute. This could possibly result in showing an attribute as an urn:mace and an urn:oid if it's present in the Assertion as both an urn:mace as an urn:oid, while other (known) attributes are deduplicated using the dictionary.

As we barely have information about the given attribute, it would be difficult to match its `urn:mace` with an `urn:oid`. Therefore, it would make sense to just show them both and pass on deduplicating them. However, this would not be consistent with the other attributes that are deduplicated.

@thijskh, what do you think?

thijskh:
Indeed, if an attribute is unknown we cannot do much optimization. So it's fine to show both `urn:mace` and `urn:oid` in case of an unknown attribute.

arothuis:
Implemented in #51 and #68

PR #68, opened 2016-10-19
Implement attribute set with fallbacks

arothuis:
This implements #55 in Profile, using the ConfigurableAttributeSetFactory from SURFnet/Stepup-saml-bundle#51.

As shown in the image below, an undefined attribute will use its urn as a name in the My Services table (in this case I removed the definition for the "surfconext status" for testing purposes). It will not throw an exception.
screen shot 2016-10-19 at 14 56 03

### DIFF
--- a/src/OpenConext/Profile/Entity/AuthenticatedUser.php
+++ b/src/OpenConext/Profile/Entity/AuthenticatedUser.php
@@ -19,11 +19,11 @@
 namespace OpenConext\Profile\Entity;
 
 use OpenConext\Profile\Assert;
-use OpenConext\Profile\Exception\InvalidEptiAttributeException;
 use OpenConext\Profile\Exception\RuntimeException;
 use OpenConext\Profile\Value\EntityId;
 use Surfnet\SamlBundle\SAML2\Attribute\Attribute;
 use Surfnet\SamlBundle\SAML2\Attribute\AttributeSet;
+use Surfnet\SamlBundle\SAML2\Attribute\AttributeSetInterface;
 use Surfnet\SamlBundle\SAML2\Response\AssertionAdapter;
 
 final class AuthenticatedUser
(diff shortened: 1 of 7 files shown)


arothuis:
@rjkip

src/OpenConext/Profile/Entity/AuthenticatedUser.php
```diff
@@ -34,7 +34,7 @@
     private $nameId;
 
     /**
-     * @var AttributeSet
+     * @var AttributeSetInterface
```
rjkip: Come to think of it, this interface, returned from the factory, is actually a BC break, innit?

src/OpenConext/ProfileBundle/Resources/views/MyServices/overview.html.twig
```diff
@@ -76,7 +76,13 @@
                                     {% endif %}
                                     <tr>
                                         <td title="{{ attributeUrn }}">
-                                            <strong>{{ ('profile.saml.attributes.' ~ attribute.attributeDefinition.name)|trans({}, 'saml') }}</strong>
+                                            {% set translatedAttribute = ('profile.saml.attributes.' ~ attribute.attributeDefinition.name)|trans({}, 'saml') %}
+
+                                            {% if translatedAttribute == 'profile.saml.attributes.' ~ attribute.attributeDefinition.name %}
```
rjkip: So... why did we prefix the attributes in the SAML translation domain with `profile.saml.attributes.`?

arothuis: I think we wanted to make them look like translation keys in order to distinguish the translation key from the raw URN.
I might rename this to `translatedAttributeName`.

rjkip: Meh, I'll allow it :hammer:

Issue #51, opened 2016-06-10
Under my-services, each service displays the same list of (wrong) attributes


thijskh:
Under my-services, each of the services I have logged into shows the same list of four attributes it supposedly received. The list should be different per SP depending of course on the attribute release policy for that SP.

DRvanR:
At the moment of creating the new Profile application this was a conscious design decision. EngineBlock does not keep track of, nor could provide a list of attributes and values exposed per SP. Therefor it would be implossible to create this functionality.

There is currently an issue in the [Profile Icebox][1] that facilitates the functionality as described. This still requires a design decision and is pending an implementation in EngineBlock (as EngineBlock should provide the data). The current implementation is created similar to the functionality as described in [this issue][2].

[1]: https://www.pivotaltracker.com/story/show/108068776
[2]: https://www.pivotaltracker.com/story/show/106201046

thijskh:
The old profile now displays the result of applying the current ARP for an SP to the current set of attributes received by profile for this user.

[108068776][1] in my reading deals with the fact that this is not strictly accurate, since the ARP, or the set of the user's attributes could have changed over time. However, it's definitely a good approximation of what happens.

I currently see the following four attributes, and it's the same for every service:
![screenshot from 2016-06-10 14 12 51](https://cloud.githubusercontent.com/assets/3808792/15963863/7134754a-2f15-11e6-8d92-bae56f6ebf12.png)
I don't understand how I would get to that list from reading [06201046][2].

[1]: https://www.pivotaltracker.com/n/projects/1453004/stories/108068776 
[2]: https://www.pivotaltracker.com/n/projects/1453004/stories/106201046

DRvanR:
The current profile is part of EngineBlock, and therefor can have the ARPs applied for the respective SPs  over all attributes it has for the current user, as EngineBlock contains the ARP logic. The new profile is not part of EngineBlock and does not contain the ARP logic (and should not), therefor Profile cannot apply the same ARPs as EngineBlock (and thus not as the current Profile). The choice to create the current implementation was discussed during the development phase of Profile.

The list that is currently shown is the result of all attributes Profile receives from EngineBlock (after applying ARP for Profile if there is any), followed by filtering [these attributes][1] as is done in that class, as was required per link in the description of https://www.pivotaltracker.com/n/projects/1453004/stories/106201046. These attributes are then indeed shown for all SPs.

Hope that explains it :smile: 

[...]

PR #70, opened 2016-10-21
Apply attribute release policy to attributes under My Services

arothuis:
This implements #51.

Relates to OpenConext/OpenConext-engineblock#350.


src/OpenConext/EngineBlockApiClientBundle/Service/AttributeReleasePolicyService.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $this->jsonApiClient = $jsonApiClient;
+    }
+
+    /**
+     * @param ConsentList $consentList
+     * @param AttributeSetInterface $attributeSet
+     * @return SpecifiedConsentList
+     */
+    public function applyAttributeReleasePolicies(ConsentList $consentList, AttributeSetInterface $attributeSet)
```
arothuis: I think this method is kinda messy. This is because we have a ConsentList and an AttributeSet and we get an array from the response that turns into a SpecifiedConsentList. We might want to find a way to separate the mapping logic from the other service logic.

src/OpenConext/EngineBlockApiClientBundle/Service/AttributeReleasePolicyService.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $entityIds = $consentList->map(function (Consent $consent) {
+            return $consent->getServiceProvider()->getEntity()->getEntityId()->getEntityId();
+        });
+
+        $mappedAttributes = [];
+        $definitions = [];
```
arothuis: This is a trick to prevent having to look up the definitions in the AttributeDictionary again: it simplifies the logic a bit.

arothuis:
@rjkip, 🐣

thijskh:
Do I understand correctly that when this is merged, Profile will depend on an EB version that supports the new API endpoint?

If so we may want to consider to release profile (say 0.3.1) without this PR merged first, so we can roll that out, and then a new release with this in it (say 0.4) which we can roll that out after...

arothuis:
Your understanding is correct. This PR (and the one for EB), still need a bit of work for robustness' sake and a bit of clean-up, but when finished and merged, EngineBlock will have a new API for allowing ARPs to be applied to a set of services and attributes. 

It makes sense to release Profile without the dependency on a new EB API first, then release a new EB version with said API and after that update and release Profile again. We should then document the dependency on a certain EB version in the requirements in the README.

thijskh:
Alright. So will you release a 0.3.1 now?

arothuis:
[0.4.0](https://github.com/OpenConext/OpenConext-profile/releases/tag/0.4.0) has been released.

src/OpenConext/EngineBlockApiClientBundle/Service/AttributeReleasePolicyService.php
```diff
@@ -0,0 +1,88 @@
...
+        ]);
+
+        return SpecifiedConsentList::createWith(
+            $consentList->map(
+                function (Consent $consent) use ($response, $definitions) {
+                    $entityId = $consent->getServiceProvider()->getEntity()->getEntityId()->getEntityId();
+
+                    $attributes = [];
+                    foreach ($response[$entityId] as $attributeName => $attributeValue) {
+                        $attributes[] = new Attribute($definitions[$attributeName], $attributeValue);
```
rjkip: It is an assumption that a definition by that attribute name was actually cached before (what if EB misbehaves?). Also, I feel like the definition caching adds another thing to keep in mind; you could also use the attribute dictionary.

Same assumption comment applies to `$response[$entityId]`. Would be nice to have included in the tests.

src/OpenConext/ProfileBundle/Service/SpecifiedConsentListService.php
```diff
@@ -54,13 +62,11 @@ public function __construct(
      */
     public function getListFor(AuthenticatedUser $user)
     {
-        $consentList        = $this->consentService->findAllFor($user);
-        $filteredAttributes = $user->getAttributes()->apply($this->attributeFilter);
```
rjkip: Can the attribute filter go?

src/OpenConext/EngineBlockApiClientBundle/Service/AttributeReleasePolicyService.php
```diff
@@ -0,0 +1,88 @@
...
+
+            $mappedAttributes[$name] = $attribute->getValue();
+
+            // Remember the attribute definitions so we can more easily build the attributes again
+            $definitions[$name] = $attribute->getAttributeDefinition();
+        }
+
+        $response = $this->jsonApiClient->post('/arp', [
+            'entityIds'  => $entityIds,
+            'attributes' => $mappedAttributes
```
rjkip: Bit of a nit, but an empty mapped attribute associative array would be encoded as `[]`, instead of `{}`. Perhaps a short-circuit of the API is not needed, as this will never happen in practice, but perhaps `!empty($mappedAttributes) ? $mappedAttributes : new stdClass()` or `(object) $mappedAttributes`.

arothuis: I know you can't always assume this will hold, but JSON's `[]` and `{}` are converted back to `[]` on the server side because we're decoding to arrays. I'm not sure about this.

src/OpenConext/ProfileBundle/Service/SpecifiedConsentListService.php
arothuis: I think we can remove the relevant classes, indeed.

src/OpenConext/EngineBlockApiClientBundle/Service/AttributeReleasePolicyService.php
rjkip: Yes, you're right. It's just a nit for correctness.

src/OpenConext/EngineBlockApiClientBundle/Http/JsonApiClient.php
```diff
@@ -83,6 +83,45 @@ public function read($path, array $parameters = [])
         return $data;
     }
 
+    public function post($resource, array $data)
```
rjkip: Not currently, but this JSON API client is not specific for the EB ARP API.

arothuis:
@rjkip, it's been a long road, are we there yet? :car:

src/OpenConext/EngineBlockApiClientBundle/Service/AttributeReleasePolicyService.php
```diff
@@ -65,10 +65,10 @@ public function applyAttributeReleasePolicies(ConsentList $consentList, Attribut
             $definitions[$name] = $attribute->getAttributeDefinition();
         }
 
-        $response = $this->jsonApiClient->post('/arp', [
+        $response = $this->jsonApiClient->post([
             'entityIds'  => $entityIds,
-            'attributes' => $mappedAttributes
-        ]);
+            'attributes' => !empty($mappedAttributes) ? $mappedAttributes : new stdClass()
+        ], '/arp');
```
rjkip: Just for readability's sake, could you put the data in a variable and then `post($data, '/arp')`?

PR #72, opened 2016-11-01
Support fallbacks for attribute definitions in ARP service

arothuis:
Because we do not have any guarantees the definitions are to be found in the dictionary, we apply the same fallback mechanism as used in the AttributeSetWithFallbacks.


thijskh:
Seems to solve the problem! :+1:

PR #85, opened 2017-12-01
Filter blacklisted attributes

MKodde:
This is a fix of the previously merged PR: #84 

Filtering the attributes from the user at earliest possible occasion was not the way to go. The SamlProvider creates an AuthenticateUser instance when the user logs in. This authenticated user is loaded with the saml attributes from the assertion of the SAML response.

Filtering the attributes at that point proved to be problematic for other parts of the application as they relied on the `urn:oid:1.3.6.1.4.1.1076.20.40.40.1` (surfconextid) to be present in the list of user attributes.

To solve this problem, a new getAttributes method was introduced on the AuthenticateUser class. Who is tasked with filtering the blacklisted attributes on a 'just in time' basis, leaving the required attributes in place for other processes to use.

### DIFF
--- a/src/OpenConext/Profile/Entity/AuthenticatedUser.php
+++ b/src/OpenConext/Profile/Entity/AuthenticatedUser.php
@@ -134,4 +134,22 @@ final class AuthenticatedUser
     {
         return $this->nameId;
     }
+
+    /**
+     * @return AttributeSet
+     */
+    public function getAttributesFiltered()
+    {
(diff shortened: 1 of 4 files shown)


thijskh:
> as they relied on the urn:oid:1.3.6.1.4.1.1076.20.40.40.1 (surfconextid) to be present
→ since this is deprecated one might consider it a problem that profile uses it. But good to leave that out of scope now. Will put it on the backlog.

PR #89, opened 2018-01-09
Show attributes from AA sources in My Services

MKodde:
**Test notes**
Enable ARP for an SP you can use (I've used teams). And set sources different than idp on some of the attributes. Log in to the service you just modified and then check the my services page in Profile. 

Update your EB installation by checking out the `feature/eb-api-read-arp-endpoint` feature branch.

A new release for the Saml bundle was released and installed (2.11.2). Make sure you update your composer dependencies.

**Work checklist**
- [x] Update EB API to be able to retrieve Arp config for SP entities
- [x] Update StepupSamlBundle to add support for the ORCiD ID attribute (Dictionary config)
- [x] Refactor my services related logic

### DIFF
--- a/src/OpenConext/Profile/Value/Arp.php
+++ b/src/OpenConext/Profile/Value/Arp.php
@@ -134,9 +134,7 @@ final class Arp
     public function getNonIdpAttributes()
     {
         $attributes = $this->getAttributesGroupedBySource();
-        if (isset($attributes['idp'])) {
-            unset($attributes['idp']);
-        }
+        unset($attributes['idp']);
         return $attributes;
     }
 
--- a/src/OpenConext/Profile/Value/Arp.php
+++ b/src/OpenConext/Profile/Value/Arp.php
@@ -1,0 +1,147 @@
+<?php
+
+/**
+ * Copyright 2017 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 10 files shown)


src/OpenConext/Profile/Value/Arp.php
```diff
@@ -0,0 +1,147 @@
...
+
+        return new self($arp, $dictionary);
+    }
+
+    private function __construct(array $arp, AttributeDictionary $dictionary = null)
+    {
+        $arpCollection = [];
+
+        // Create Attribute instances for all attributes
+        foreach ($arp as $attributeName => $attributeDefinitionInformation) {
```
jorissteyn: Can't we lazy-load this?

src/OpenConext/Profile/Value/Arp.php
```diff
@@ -0,0 +1,147 @@
...
+        return true;
+    }
+
+    /**
+     * @return array
+     */
+    public function getNonIdpAttributes()
+    {
+        $attributes = $this->getAttributesGroupedBySource();
+        if (isset($attributes['idp'])) {
```
jorissteyn: This is redundant: just unset(..) will do :)

src/OpenConext/Profile/Value/Arp.php
MKodde: Can be done but is quite impractical to do so at this moment.

src/OpenConext/Profile/Value/Arp.php
MKodde: TIL

PR #100, opened 2018-03-02
Add custom error templates for 403, 404 and 500 errors

jorissteyn:
See: https://www.pivotaltracker.com/story/show/155549384

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,10 @@ can be done on an installation that runs in `DEV` mode. Make sure you log
 into profile at `https://profile-dev.vm.openconext.org`. Then go to
 `https://profile.vm.openconext.org/app_dev.php/_trans/` to update the strings.
 
+The following command can be used to scan for translations:
+
+     ./bin/extract-translations.sh
+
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -94,6 +94,10 @@ can be done on an installation that runs in `DEV` mode. Make sure you log
 into profile at `https://profile-dev.vm.openconext.org`. Then go to
 `https://profile.vm.openconext.org/app_dev.php/_trans/` to update the strings.
 
+The following command can be used to scan for translations:
+
+     ./app/console translation:extract --output-dir=app/Resources/translations/ --dir=src --dir=app/Resources
```
MKodde: This is valuable information. It might be better to advice the user to use the `bin/extract-translations.sh` script to achieve this goal?

MKodde: The output generated between your way of extracting the translations and the way `bin/extract-translations.sh` achieves this results in many changes in the translation files (`jms:reference-file` paths are appended with four `../` unnecessarily).

PR #101, opened 2018-03-02
Don't show transient and persistent name IDs

jorissteyn:
This requires EB5.5.2 after hotfix1, or EB>5.5.4 or EB>5.6.3 or EB>=5.7.

Quick summary of how this works:

 - EB sends the configured NameIDFormat of the SP in the response of the /consent API call
 - profile shows the italic 'transient' message instead of the eduPersonTargetedID if the format is transient
 - profile shows the italic 'persistent' message instead of the eduPersonTargetedID if the format is persistent
 - values which are not transient or persistent nameIDs (unspecified, email, ...) are shown in the attributes table as-is

See: https://www.pivotaltracker.com/story/show/154912300
And: https://github.com/OpenConext/OpenConext-engineblock/pull/519


jorissteyn:
Travis seems down, let's see if the build passes next week.

PR #105, opened 2018-05-24
Remove duplicate 'support_url' in consent API response

jorissteyn:
EB previously sent the support URL in the 'support_url' and
'support_url_*' fields, the latter was removed in EB5.7 and this
commit updates profile to support that change.

See https://github.com/OpenConext/OpenConext-engineblock/pull/532

### DIFF
--- a/src/OpenConext/EngineBlockApiClientBundle/Value/ConsentListFactory.php
+++ b/src/OpenConext/EngineBlockApiClientBundle/Value/ConsentListFactory.php
@@ -109,8 +109,7 @@ final class ConsentListFactory
         Assert::keyExists($data, 'eula_url', 'Consent JSON structure must contain key "eula_url"');
         Assert::keyExists($data, 'support_email', 'Consent JSON structure must contain key "support_email"');
         Assert::keyExists($data, 'name_id_format', 'Consent JSON structure must contain key "name_id_format"');
-        Assert::keyExists($data, 'support_url_en', 'Consent JSON structure must contain key "support_url_en"');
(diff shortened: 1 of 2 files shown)


src/OpenConext/EngineBlockApiClientBundle/Value/ConsentListFactory.php
```diff
@@ -109,8 +109,7 @@ private static function createServiceProvider($data)
         Assert::keyExists($data, 'eula_url', 'Consent JSON structure must contain key "eula_url"');
         Assert::keyExists($data, 'support_email', 'Consent JSON structure must contain key "support_email"');
         Assert::keyExists($data, 'name_id_format', 'Consent JSON structure must contain key "name_id_format"');
-        Assert::keyExists($data, 'support_url_en', 'Consent JSON structure must contain key "support_url_en"');
-        Assert::keyExists($data, 'support_url_nl', 'Consent JSON structure must contain key "support_url_nl"');
+        Assert::keyExists($data, 'support_url', 'Consent JSON structure must contain key "support_url"');
```
MKodde: Could you not be even more specific here? Yes, the support_url key must exist but it should also be an array, with one of keys: `en` or `nl`.

MKodde: Above mentioned situation could easily be tested in the ConsentListFactoryTest

jorissteyn: In practice, yes, but in theory, maybe it only contains the keys 'fr' and 'es'. Should still work, it only uses en/nl when present.

PR #107, opened 2018-06-21
Include arp motivations

MKodde:
This PR changes two things

1. The my services screen no longer breaks when attribute motivations are returned from the `read-arp` engine block API endpoint. This was fixed by making the response validation more tolerant. Now only the value entry is actively tested. The source and motivation are optional.
2. The motivation is displayed in the 'my services' overview.

The Pivotal story: https://www.pivotaltracker.com/story/show/158522286

### DIFF
--- a/src/OpenConext/Profile/Value/Arp.php
+++ b/src/OpenConext/Profile/Value/Arp.php
@@ -122,16 +122,15 @@ final class Arp
     private static function isValidAttribute(array $attributeInformation)
     {
         foreach ($attributeInformation as $attributeInformationEntry) {
-            if (!array_key_exists('value', $attributeInformationEntry)) {
+            if (!isset($attributeInformationEntry['value'])) {
                 return false;
             }
 
-            foreach ($attributeInformationEntry as $key => $value) {
+            foreach ($attributeInformationEntry as $value) {
                 if (!is_string($value)) {
                     return false;
                 }
             }
-
         }
         return true;
     }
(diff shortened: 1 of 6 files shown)


MKodde:
The build breaks on the security checker. I'm opening a new PR with the upgrade of Symfony. Fixing this issue.

MKodde:
#108 was added to this PR

src/OpenConext/Profile/Value/Arp.php
```diff
@@ -142,4 +149,37 @@ public function getAttributesGroupedBySource()
...
+    }
+
+    public function getMotivationFor(Attribute $attribute)
+    {
+        foreach ($this->arp as $arpSource) {
+            foreach ($arpSource as $arpEntry) {
+                if ($attribute->getAttributeDefinition()->getUrnMace() == $arpEntry->getAttributeDefinition()->getUrnMace()) {
+                    $values = $arpEntry->getValue();
+                    $attributeValue = reset($values);
+                    if (array_key_exists('motivation', $attributeValue)) {
```
jorissteyn: Do you really want to allow null values here?

I believe isset() in most cases is more idiomatic, array_key_exists() is usually only used when you explicitly want to allow null values.

jorissteyn: Applies to the two other usages of array_key_exists() to.

src/OpenConext/ProfileBundle/Resources/views/MyServices/AttributeList/idp.html.twig
```diff
@@ -1,10 +1,13 @@
 <p>{{ 'profile.my_services.attribute_release_description'|trans }}</p>
-
+{% set showMotivations = specifiedConsent.hasMotivations %}
```
jorissteyn: Might be simpler just checking hasMotivations twice instead of assigning variables inside the template, what do you think?

src/OpenConext/ProfileBundle/Resources/views/MyServices/AttributeList/idp.html.twig
```diff
@@ -1,10 +1,13 @@
...
 <table class="mdl-data-table">
     <thead>
     <tr>
         <th>{{ 'profile.table.attribute_name'|trans }}</th>
         <th>{{ 'profile.table.attribute_value'|trans }}</th>
+        {% if showMotivations %}
+            <th>{{ 'profile.table.motivation'|trans }}</th>
```
jorissteyn: Is 'motivation' what we want to call it to end users? Maybe take a screenshot and hang it in pivotal for surf to bikeshed.

MKodde:
Thanks for the feedback. I've amended some fixes to the commits and added a screenshot to Pivotal.

PR #111, opened 2018-12-13
Log which URL failed validation.

thijskh:
Makes it a bit less of a wild goose chase to find the one invalid URL in
the database.

### DIFF
--- a/src/OpenConext/Profile/Value/Url.php
+++ b/src/OpenConext/Profile/Value/Url.php
@@ -18,7 +18,7 @@
 
 namespace OpenConext\Profile\Value;
 
-use OpenConext\Profile\Exception\InvalidArgumentException;
+use OpenConext\Profile\Assert;
 
 final class Url
 {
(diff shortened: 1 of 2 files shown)


src/OpenConext/Profile/Value/Url.php
```diff
@@ -33,7 +33,7 @@
     public function __construct($url)
     {
         if (filter_var($url, FILTER_VALIDATE_URL) === false) {
-            throw new InvalidArgumentException('URL must be valid');
+            throw new InvalidArgumentException('URL must be valid: "' . $url . '"');
```
MKodde: Just personal preference, but I always prefer the use of `sprintf` in these situations.

`throw new InvalidArgumentException(sprintf('URL must be valid: ''%s"', $url));`

But don't sweat it!

src/OpenConext/Profile/Value/Url.php
```diff
@@ -32,9 +32,8 @@
      */
     public function __construct($url)
     {
-        if (filter_var($url, FILTER_VALIDATE_URL) === false) {
-            throw new InvalidArgumentException('URL must be valid: "' . $url . '"');
-        }
+        Assert::string($url, 'URL "%s" must be a string');
```
MKodde: Had to look twice, I think you can drop the first commit of this PR :)

Nice proposal, you've got my approval.

PR #113, opened 2019-02-20
Security updates

MKodde:
Fix the security warnings that are reported by the security checker.

Note that we use the SimpleSAMLphp SAML2 library at version 3.2 until 3.3 is considered stable.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,8 +31,7 @@
         "nelmio/security-bundle": "^1.8",
         "symfony/swiftmailer-bundle": "^2.4",
         "openconext/monitor-bundle": "^1.0",
-        "paragonie/random_compat": "2.0.11 as 1.4.2",
-        "simplesamlphp/saml2": "3.2.*"
+        "paragonie/random_compat": "2.0.11 as 1.4.2"
     },
     "require-dev": {
         "sensio/generator-bundle": "~2.3",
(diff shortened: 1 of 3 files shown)


MKodde:
Interesting, the failing test is broken because of a SAML2 3.2 feature which I am yet to find.
https://github.com/OpenConext/OpenConext-profile/blob/b4a44eee83aeb2cca81f3503315b7e35743d34aa/src/OpenConext/Profile/Tests/Entity/AuthenticatedUserTest.php#L141-L154

The test works with SAML2 @ 3.1.6 but breaks on 3.2.6.

MKodde:
Ive been researching the problem. And the SAML2 lib used to have a very explicit check on the correctness of the EPTI attribute. This was added in: 

https://github.com/simplesamlphp/saml2/commit/d3b8bb0d7d15f0ff0b3eba249d1d2d1a5f745c8b  (https://github.com/simplesamlphp/saml2/pull/60 and later https://github.com/simplesamlphp/saml2/pull/76)

Then @thijskh made the validation less harsh, by logging the problem instead of throwing an Exception. in this commit: https://github.com/simplesamlphp/saml2/commit/6db983648ca2607bf0de8025f30f3007e962c374 which landed in v3.2.3.

In order to be compatible with versions >= v3.2.3 we need to update the test or make SAML2 more strict again. 

I'm opting for the first option. 

@thijskh can you remember the reason for such a specific test case in the AuthenticatedUserTest.php? Is it something that happens often, having an assertion with more than one EPTI values?

thijskh:
That should really never happen. Does the commit history not tell anything about it? Otherwise safe to drop it.

MKodde:
Nope no insights to be found in gitlog. The test is actually more a test of the SAML bundle. Removing the test.

PR #114, opened 2019-02-21
Stop using SURFconextId as user identifier

MKodde:
The nameId value is now used instead. Containing the exact same value
as long as the NameID was configured to be unspecified (which it is in
our case).

The NameID can already be read from the AuthenticatedUser object, so
manually retrieving it from the assertion subject is not required.

The ConsentService could be cleaned up drastically.

https://www.pivotaltracker.com/story/show/164103691

### DIFF
--- a/src/OpenConext/ProfileBundle/Service/ConsentService.php
+++ b/src/OpenConext/ProfileBundle/Service/ConsentService.php
@@ -27,28 +27,13 @@ use Surfnet\SamlBundle\SAML2\Attribute\AttributeDefinition;
 final class ConsentService
 {
     /**
-     * @var LoggerInterface
-     */
-    private $logger;
-
-    /**
      * @var ConsentRepository
      */
     private $consentRepository;
 
-    /**
-     * @var AttributeDefinition
-     */
-    private $identifyingAttribute;
-
-    public function __construct(
-        LoggerInterface $logger,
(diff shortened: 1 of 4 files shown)


MKodde:
The build breaks on the security checker. These issues are addressed in #113

src/OpenConext/ProfileBundle/Service/ConsentService.php
```diff
@@ -57,34 +42,6 @@ public function __construct(
...
-                $this->identifyingAttribute->getName()
-            );
-
-            $this->logger->error($message);
-
-            return null;
-        }
-
-        return $this->consentRepository->findAllFor($identifyingValue);
+        return $this->consentRepository->findAllFor($user->getNameId());
```
pablothedude: Shouldn't there be any validation on the nameId? Because the AssertionAdapter could return null as value and maybe we want to log that.

MKodde: `\OpenConext\Profile\Entity\AuthenticatedUser::__construct` asserts that the NameId is a string. There could be a very rare case that the NameID value is an empty string. Then, the `\OpenConext\Profile\Repository\ConsentRepository::findAllFor` method could be called with an empty string. But I don't see that happening.

PR #117, opened 2019-03-15
Portuguese translations

domgon:
I made some adaptations do adapt the Portuguese language to the Profile. On the translations.html.twig I add the locale for Portuguese local at the end to trying to avoid problems with the lines because of the messages on the saml.[en,nl,pt].xliff messages.


app/config/config.yml
```diff
@@ -5,7 +5,7 @@ imports:
     - { resource: global_view_parameters.yml }
 
 parameters:
-    locales: [en, nl]
+    locales: [en, nl, pt]
```
MKodde: Is this OK with you @thijskh? Adding PT to the list of locales will show the Portuguese language in the language switcher. Overriding the locales from the `parameters.yml` does not work. 

Two options:
1. Leave it as is, and we include this feature for our own Portuguese speaking users
2. Move the locale parameter definition to the parameters.yml and remove it from the config.yml, allowing configuration of the locale switcher from the parameters.yml.

app/config/global_view_parameters.yml.dist
```diff
@@ -5,15 +5,20 @@ parameters:
...
         nl: "https://support.surfconext.nl/terms-nl"
+        pt: "https://support.surfconext.nl/terms-en"
     platform_url:
         en: "https://www.surfconext.nl/en"
         nl: "https://www.surfconext.nl"
+        pt: "https://www.surfconext.nl/en"
     profile_explanation_image_path:
         en: "/images/profile_home_en.png"
         nl: "/images/profile_home_nl.png"
+        pt: "/images/profile_home_pt.png"
```
MKodde: Nice job, updating the PNG!

app/config/config.yml
domgon: Thank you guys.
When I see some spelling or grammar in Portuguese that is not in the best way I will correct and submit the changes.

thijskh: 2

MKodde: Then let's merge this PR and I will make the config configurable in a new PR

PR #126, opened 2019-06-20
Upgrade to Symfony 3.4 and fix deprecation issues while at it

MKodde:
This PR is way too big. It is very hard to make an upgrade like this more structured and thus easier to review. The four commits that make up this PR can be reviewed step for step, making it (hopefully) slightly easier.

The main exercise was to upgrade to Symfony 3.4. This was followed by a boatload of incompatibility fixes and an attempt to align more closely with the Symfony 3/4 way of configuring a project. Next, deprecated features have been removed from the project when possible. Some warnings remain due to 3rd party deprecation warnings.

Finally some housekeeping was required to get the Travis train back on tracks. Some highlights:
- Explicitly install `ant`
- Added a 'high coupling' warning suppression

And some other minor improvements:
- Move the PHPUnit config from `app/` to the project root
- Removed some RMT leftovers

See: https://www.pivotaltracker.com/story/show/164138880


app/autoload.php
```diff
@@ -1,29 +1,8 @@
 <?php
 
-/**
```
BackEndTea: All this file now does is require `vendor/autoload`. How about just removing it and requiring vendor/autoload directly where this file is required now.

bin/console
```diff
@@ -0,0 +1,23 @@
...
+<?php
+
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Debug\Debug;
+
+set_time_limit(0);
+
+/** @var Composer\Autoload\ClassLoader $loader */
+$loader = require __DIR__.'/../app/autoload.php';
```
BackEndTea: No need to wire the loader to a variable, as it is never used

src/OpenConext/AttributeAggregationApiClientBundle/Resources/config/services.yml
```diff
@@ -1,18 +1,9 @@
...
         arguments:
-            - @openconext_aa_api.guzzle
-
-    openconext_aa_api.repository.attribute_aggregation:
-        public: false
-        class: OpenConext\AttributeAggregationApiClientBundle\Repository\AttributeAggregationRepository
-        arguments:
-            - @openconext_aa_api.service.api
\ No newline at end of file
+            - '@openconext_aa_api.guzzle'
```
BackEndTea: Missing a new line at the end of the file.

This happens in a few other places as well. How about adding an `.editorconfig` in another PR to reduce these kinds of errors?

For more info see [their website](https://editorconfig.org/)

src/OpenConext/ProfileBundle/DependencyInjection/OpenConextProfileExtension.php
```diff
@@ -126,7 +127,7 @@ private function parseLocaleCookieStorageConfiguration(
         }
 
         $container
-            ->getDefinition('profile.storage.locale_cookie')
+            ->getDefinition('OpenConext\ProfileBundle\Storage\SingleCookieStorage')
```
BackEndTea: How about using the `::class` notation here?

src/OpenConext/ProfileBundle/DependencyInjection/OpenConextProfileExtension.php
```diff
@@ -137,7 +138,7 @@ private function parseLocaleCookieStorageConfiguration(
     private function parseEngineBlockAttributeAggregationConfiguration($aaConfig, ContainerBuilder $container)
     {
         $container
-            ->getDefinition('profile.available_aa_attributes')
+            ->getDefinition('OpenConext\Profile\Value\AttributeAggregation\AttributeAggregationEnabledAttributes')
```
BackEndTea: How about using the `::class` notation here?

src/OpenConext/ProfileBundle/Form/Type/SwitchLocaleType.php
```diff
@@ -104,6 +106,6 @@ private function formatLocaleChoices(LocaleSet $availableLocales)
             $localeChoices[$locale->getLocale()] = 'profile.locale.' . $locale->getLocale();
         }
 
-        return $localeChoices;
+        return array_flip($localeChoices);
```
BackEndTea: What is the reason for flipping this array now? 

What about switching the keys and values when assigning it in the for each loop above?

.travis.yml
```diff
@@ -1,13 +1,11 @@
 language: php
 
 php:
-  - 5.6
```
BackEndTea: This to me indicates we are dropping php 5.6 support.

If that is the case, then how about removing it from the `composer.json` as well?

If we are not yet dropping 5.6 support, then how about we keep testing it, untill we are 100% sure that support is dropped.

src/OpenConext/Profile/Api/AuthenticatedUserProviderInterface.php
```diff
@@ -20,7 +20,7 @@
 
 use OpenConext\Profile\Entity\AuthenticatedUser;
 
-interface AuthenticatedUserProvider
+interface AuthenticatedUserProviderInterface
```
BackEndTea: This PR contains a big amount of classes(interfaces) that simply had their name changed. This makes it a bit harder to find exactly what is changed.

Is this because a coding style updated?

app/AppKernel.php
```diff
@@ -60,6 +54,21 @@ public function registerBundles()
...
     }
 
+    public function getRootDir()
+    {
+        return __DIR__;
+    }
+
+    public function getCacheDir()
+    {
+        return dirname(__DIR__).'/var/cache/'.$this->getEnvironment();
```
BackEndTea: If i remember correctly, `__DIR__ . '/../` is a bit faster than  `dirname(__DIR__)`, because it can be evaluated at compile time.

app/autoload.php
```diff
@@ -1,29 +1,8 @@
...
- */
-
-use Doctrine\Common\Annotations\AnnotationRegistry;
 use Composer\Autoload\ClassLoader;
 
-/**
- * @var ClassLoader $loader
- */
+/** @var ClassLoader $loader */
 $loader = require __DIR__.'/../vendor/autoload.php';
```
BackEndTea: No need to write the loader to a variable, we can return it directly

app/autoload.php
MKodde: Good point, I'm removing the app/autoload.php and will use the composer autoload file directly instead.

app/autoload.php
MKodde: Indeed, removed the declaration

src/OpenConext/Profile/Api/AuthenticatedUserProviderInterface.php
MKodde: This is done to allow the Symfony autowire feature to work a little better. Having classes/interfaces with the same name in different namespaces make it harder to get it to work.

.travis.yml
MKodde: This version of Profile will only work on PHP 7.0, so removing the support in the composer.json is a good suggestion.

web/app.php
```diff
@@ -1,14 +1,18 @@
 <?php
 
-use Symfony\Component\ClassLoader\ApcClassLoader;
 use Symfony\Component\HttpFoundation\Request;
 
-$loader = require_once __DIR__.'/../app/bootstrap.php.cache';
+require __DIR__.'/../vendor/autoload.php';
 
-require_once __DIR__.'/../app/AppKernel.php';
+if (PHP_VERSION_ID < 70000) {
```
BackEndTea: Since we only support 7.0 or higher, these if statements can be removed

src/OpenConext/ProfileBundle/Form/Type/SwitchLocaleType.php
MKodde: Yes, one operation less to deal with. 

While adding the flip function, my consideration was that it looked slightly better, not having the string concatenation in the array key.

src/OpenConext/AttributeAggregationApiClientBundle/Resources/config/services.yml
MKodde: Added the newline, but not yet added an editorconfig.

web/app.php
MKodde: Good point, updated this!

PR #129, opened 2019-06-24
Correctly capitalize alt text

BackEndTea:
ORCID ID logo instead of ORCID iD logo


BackEndTea:
My bad. Guess some changes in @MKodde his last PR also need updating then.

PR #130, opened 2019-06-25
Install and configure Webpack Encore

MKodde:
The individual commits hint at the changes required to install and configure Encore at a more atomic level.

See: https://www.pivotaltracker.com/story/show/166868666


web/js/app.js
```diff
@@ -0,0 +1,9 @@
+require('jquery');
```
BackEndTea: The standard nowdays is to use import

web/js/app.js
```diff
@@ -0,0 +1,9 @@
...
+require('../../node_modules/material-design-lite/dist/material.min.js');
+
+require('./main.js');
+
+require('../../node_modules/material-design-lite/dist/material.min.css');
```
BackEndTea: These shouldn't be relative to this, but instead something like: `material-design-lite/dist/material.min.css`

web/js/app.js
```diff
@@ -0,0 +1,9 @@
+require('jquery');
+require('../../node_modules/material-design-lite/dist/material.min.js');
```
BackEndTea: Since the `material-design-lite` package.json specifies to import this file, `'material-design-lite' should be enough here.

package.json
```diff
@@ -0,0 +1,28 @@
+{
```
BackEndTea: Encore does a lot of work under the hood for us. This includes compiling down to certain js versions if i remember correctly.

We can configure this ourselves by adding a browserlist configuration in the package.json. It may be worthwile to find out what the defaults are, and if those align with the needs of this project.

webpack.config.js
```diff
@@ -0,0 +1,29 @@
...
+        from: './web/images/',
+        to: 'images/[path][name].[hash:8].[ext]',
+    })
+
+    .addEntry('app', './web/js/app.js')
+    .addEntry('my-connections', './web/js/my-connections.js')
+
+    // will require an extra script tag for runtime.js
+    // but, you probably want this, unless you're building a single-page app
+    .disableSingleRuntimeChunk()
```
BackEndTea: Is there a reason to disable the single runtime chunk?

src/OpenConext/ProfileBundle/Resources/views/layout.html.twig
```diff
@@ -2,15 +2,9 @@
 
 {% block stylesheets %}
     <!--[if IE]>
-    <script src="{{ asset('js/es5-shim.min.js') }}"></script>
-    <script src="{{ asset('js/classList.min.js') }}"></script>
-    <script src="{{ asset('js/selectivizr.min.js') }}"></script>
+    <script src="{{ asset('build/app.js') }}"></script>
```
BackEndTea: Will we run into issues on IE if we include build/app.js twice?

.gitignore
```diff
@@ -16,6 +16,7 @@ cache.properties
 !var/sessions/.gitkeep
 /var/bootstrap.php.cache
 !var/SymfonyRequirements.php
-/build/
+/web/build
```
BackEndTea: Personally, i like to add a `/` at the end here to indicate that this is a directory. What do you think?

src/OpenConext/ProfileBundle/Resources/views/layout.html.twig
```diff
@@ -91,7 +85,5 @@
 {% endblock %}
 
 {% block javascripts %}
-    <script src="{{ asset('js/material.min.js') }}"></script>
-    <script src="{{ asset('js/jquery-3.4.1.min.js') }}"></script>
-    <script src="{{ asset('js/main.js') }}"></script>
+    <script src="{{ asset('build/app.js') }}"></script>
```
BackEndTea: Encore has some nice helper functions for the includes, it may be better to use those

composer.json
```diff
@@ -24,6 +24,7 @@
         "symfony/monolog-bundle": "^3.3",
         "symfony/swiftmailer-bundle": "^2.4",
         "symfony/symfony": "3.4.*",
+        "symfony/webpack-encore-bundle": "^1.6",
```
BackEndTea: This bundle comes with some configuration files that seem to be missing in this PR.
How about adding those, so we can use the power that this bundle provides?

webpack.config.js
MKodde: Yes, it prevents jquery from being loaded globally, and I was not able to fix this in an elegant manner. This was the proposed fix I read on a Symfony issue.

src/OpenConext/ProfileBundle/Resources/views/layout.html.twig
MKodde: Ah, yes this is a leftover of the IE9 support we dropped. Removing this!

web/js/app.js
MKodde: For now I'm staying at requiring the packages instead of importing them.

web/js/app.js
MKodde: Updated this

web/js/app.js
MKodde: This dependency was not found: `material-design-lite in ./web/js/app.js` maybe with an import this would work. Leaving it as it is now.

package.json
MKodde: Thanks for the reminder. I offloaded this to a new ticket. For now supporting ancient browsers is not a requirement. And the out of the box settings are probably good enough.

BackEndTea:
The files that encore builds are still in the web folder, how about moving those outside of that folder so they can't be accessed over the web anymore, as that is no longer intended?

composer.json
MKodde: Can you be more specific. I know there are more encore settings to be configured in the `webpack.config.js` file. But the current setup seems to be pretty good for this project.

BackEndTea: The bundle has its own configuration: https://github.com/symfony/webpack-encore-bundle#configuration

This is required for the helper functions that come with the bundle to work

web/js/app.js
BackEndTea: That is strange, as it does work for jquery.

MKodde:
Thanks for your review @BackEndTea. Some additional feedback:

> The files that encore builds are still in the web folder, how about moving those outside of that folder so they can't be accessed over the web anymore, as that is no longer intended?

They are removed from the build artefact when that is constructed. It's a good suggestion to move them to another folder. Will add a chore in our issue tracker for that.

And regarding:

> This bundle comes with some configuration files that seem to be missing in this PR. How about adding those, so we can use the power that this bundle provides?

&

> Encore has some nice helper functions for the includes, it may be better to use those

For now leaving my setup as is, my aim was to lever Encore with the minimum amount of changes to the application. And to start using the cool new features when we released this version.

.travis.yml
```diff
@@ -20,8 +20,8 @@ before_script:
   - sudo apt-get install ant
   - phpenv config-add travis.php.ini
   - composer self-update
-  - if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.6" ]]; then composer install --prefer-dist -n -o; fi
-  - if [[ ${TRAVIS_PHP_VERSION:0:3} != "5.6" ]]; then composer install --prefer-dist --ignore-platform-reqs -n -o; fi
+  - composer install --prefer-dist --ignore-platform-reqs -n -o
```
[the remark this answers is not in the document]
MKodde: Still removed that piece, now that we drop PHP 5.6 support, that is no longer required.

PR #133, opened 2019-08-30
Add ECK ID saml attribute translation

pablothedude:
The translation was missing for the Saml attribute. This is now fixed.

https://www.pivotaltracker.com/story/show/167191228

### DIFF
--- a/app/Resources/translations/messages.pt.xliff
+++ b/app/Resources/translations/messages.pt.xliff
@@ -1,25 +1,25 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
-  <file date="2018-06-21T16:11:45Z" source-language="en" target-language="pt" datatype="plaintext" original="not.available">
+  <file date="2019-08-30T09:34:31Z" source-language="en" target-language="pt" datatype="plaintext" original="not.available">
     <header>
       <tool tool-id="JMSTranslationBundle" tool-name="JMSTranslationBundle" tool-version="1.1.0-DEV"/>
-      <note>O nó de origem, na maioria dos casos, contém a mensagem de exemplo conforme escrita pelo programador. Se parecer um ponto como delimitador como "form.label.firstname", então o programador não forneceu a mensagem por defeito.</note>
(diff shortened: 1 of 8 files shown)


app/Resources/translations/messages.pt.xliff
```diff
@@ -1,25 +1,25 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
-  <file date="2018-06-21T16:11:45Z" source-language="en" target-language="pt" datatype="plaintext" original="not.available">
+  <file date="2019-08-30T09:34:31Z" source-language="en" target-language="pt" datatype="plaintext" original="not.available">
     <header>
       <tool tool-id="JMSTranslationBundle" tool-name="JMSTranslationBundle" tool-version="1.1.0-DEV"/>
-      <note>O nó de origem, na maioria dos casos, contém a mensagem de exemplo conforme escrita pelo programador. Se parecer um ponto como delimitador como "form.label.firstname", então o programador não forneceu a mensagem por defeito.</note>
+      <note>The source node in most cases contains the sample message as written by the developer. If it looks like a dot-delimitted string such as "form.label.firstname", then the developer has not provided a default message.</note>
```
MKodde: Is this an intentional revert from PT to EN?

pablothedude: This is no revert but is set by the JMS translation bundle when running the `extract-translations.sh` script and doesn't seem to be used. I think this wasn't executed when the PT translations were added while according to the documentation should be done.

pablothedude:
Some boyscouting was done to update all translations this was done to be able to handle all translations. I've added this to the commit message.

app/Resources/translations/messages.pt.xliff
MKodde: So it's okay to lose this translation? If not used, I would expect the extractor to remove the translation.

MKodde:
> Some boyscouting was done to update all translations this was done to be able to handle all translations. I've added this to the commit message.

Nice! ready to merge. After maybe reverting the PT text mentioned above.

app/Resources/translations/messages.pt.xliff
pablothedude: It's no translation but a comment that gets filled when running the `translation:extract` command from Symfony so it wouldn't change after this change anymore.

PR #135, opened 2019-08-30
Ansible provisioning

pablothedude:
It seemed impossible to throw away the vagrant box and to provision the
box from scratch. Also npm and composer were not in place. So now when
provisioning the base box it all should work.


MKodde:
Tried your changes. And all seemed well :+1:

PR #137, opened 2019-09-16
Fix npm security issues

pablothedude:
Fix Npm security issues and add a newer version of Nodejs and Ant
to the development Ansible to be able to run the audit tools in our
development environment. A newer version of Npm was needed to be able
to run `npm audit fix`.

https://www.pivotaltracker.com/story/show/168473817

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -2,7 +2,7 @@
 
 <project name="Profile-ci" default="build">
     <target name="build"
-            depends="php-lint-ci,phpmd-ci,phpcs-ci,phpcpd-ci,phpunit-ci,php-security-checker"/>
+            depends="php-lint-ci,phpmd-ci,phpcs-ci,phpcpd-ci,phpunit-ci,php-security-checker,npm-audit"/>
 
     <target name="get-changeset.php.raw"
             description="creates a list of changed php files separated by newline">
@@ -93,6 +93,12 @@
         <exec executable="vendor/bin/phpunit" failonerror="true">
             <arg line="--configuration=${basedir}/phpunit.xml"/>
             <arg line="--coverage-text"/>
+        </exec>
+    </target>
+
+    <target name="npm-audit" description="Run npm audit">
+        <exec executable="npm" failonerror="true">
+            <arg line="audit"/>
         </exec>
     </target>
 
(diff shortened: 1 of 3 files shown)


build.xml
```diff
@@ -2,7 +2,7 @@
 
 <project name="Profile-ci" default="build">
     <target name="build"
-            depends="php-lint-ci,phpmd-ci,phpcs-ci,phpcpd-ci,phpunit-ci,php-security-checker"/>
+            depends="php-lint-ci,phpmd-ci,phpcs-ci,phpcpd-ci,phpunit-ci,php-security-checker,npm-audit"/>
```
MKodde: May I suggest merging `php-security-checker` and `npm-audit` in a single `security-check` task. As we do in most other projects? This works fine too, but running them together is even better in my opinion.

package-lock.json
```diff
@@ -883,7 +883,7 @@
         "source-map": {
           "version": "0.6.1",
           "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.6.1.tgz",
-          "integrity": "sha512-UjgapumWlbMhkBgzT7Ykc5YXUT46F0iKu8SGXq0bcwP5dz/h0Plj6enJqjz1Zbq2l5WaqYnrVbwWOWMyF3F47g==",
+          "integrity": "sha1-dHIq8y6WFOnCh6jQu95IteLxomM=",
```
MKodde: Next time, please commit the result of the audit --fix in a separate commit. This way you can describe which package needed an upgrade and why. This information is now lost.

pablothedude: I found the change to little to document, mainly because this was in Lodash and this isn't a dependency we use ourselves directly.

build.xml
pablothedude: For this single task I couldn't find any added befit to run them both at once. Because you normally want to run the checkers for npm or composer separately or you want to run all the tests together. Would you please elaborate why you want to run all security checks together so I can make a better decision in the future? Thanks!

package-lock.json
MKodde: I still think keeping commits atomic is worthwhile. For example your commit: 

1. Addresses the security issue
2. Installs yarn in able to run the tests
3. Enables the npm audit task

Three very much related tasks, but individual tasks nonetheless.

build.xml
MKodde: You can keep them as separate tasks, but providing a task that combines them makes it more convenient to run them in your dev env. Basically the same reason why we bundle the different static analysis tasks.

PR #140, opened 2019-11-07
Replace Ant with Phing

pablothedude:
In order to drop the dependency on Java Phing should be used
instead of Ant. Currently Travis has no default support for Ant.


pablothedude:
This was a test and because it didn't work out-of-the-box it was closed

PR #171, opened 2021-06-22
Upgrade stepup/saml-bundle to latest version

Badlapje:
Prior to this change, we were three versions behind on the stepup/saml-bundle

This change updates the bundle to the latest version.

Pivotal ticket @ https://www.pivotaltracker.com/story/show/175161599

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1930,16 +1930,16 @@
         },
         {
             "name": "surfnet/stepup-saml-bundle",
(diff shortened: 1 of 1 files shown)


pablothedude (approved):
Nice!

thijskh (commented):
This is already an upgrade, but I'd want us to skip to 4.2 after we've bumped the SF version of Profile to 4.4.

PR #172, opened 2021-06-22
Remove support for filtering on EB entity ID

Badlapje:
Prior to this change, the filtering was needed for the generic "these entity ids should be filtered problem".  Given that the special cased entity has been removed from EB, this is no longer needed.

This change removes the filtering.

Pivotal ticket @ https://www.pivotaltracker.com/story/show/111038134

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -129,7 +129,6 @@ open_conext_attribute_aggregation_api_client:
         verify_ssl: '%attribute_aggregation_api_verify_ssl%'
 
 open_conext_profile:
-    engine_block_entity_id: '%saml_remote_idp_entity_id%'
     locales: '%locales%'
     default_locale: '%default_locale%'
     locale_cookie_domain: '%open_conext_locale_cookie_domain%'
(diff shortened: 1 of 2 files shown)


thijskh (changes requested):
Do I now only see the removal of the parameter, not of the actual filtering?

Badlapje:
> Do I now only see the removal of the parameter, not of the actual filtering?

As far as i can see the `parseEngineBlockEntityIdConfiguration` method is the filtering?  If not I'm missing something.  I don't see anything else happening involving the engineblock entity ID.  

With regards to filtering: there's an AttributeFilter class, but that doesn't seem to do anything with the entityID?

thijskh (changes requested):
I am more confused now than I was before this PR. It seems that the engine block entity ID is also used in the process to determine the user's actual IdP (walk AuthenticatingAuthority and exclude EB). So we cannot drop the configuration. I cannot find that Profile currently does something specific with excluding the EB from the consent iist at all. So maybe we can just close this task altogether :woman_shrugging:

Badlapje:
@MKodde & @pablothedude do you guys have any insights on this?  This requires more knowledge of the inner mechanics of EB & Profile than i possess at the moment.

thijskh:
Let's just close it. No value to be gained it seems.

PR #174, opened 2021-06-28
Upgrade symfony to lts

Badlapje:
Prior to this change, the Symfony version was at 3.4.28.  This was already several versions behind the latest 3.4 version (3.4.49), but 3.4 is down to security fixes only.

This change upgrades Symfony to a LTS version (either 4.4.25 or 5.3.2).

Pivotal ticket @: https://www.pivotaltracker.com/story/show/177453745


docker/conf/xdebug.ini
```diff
@@ -1,4 +1,3 @@
 xdebug.mode = debug
 xdebug.discover_client_host = true
-xdebug.start_with_request = yes
```
quartje, line 3: Less config, less clutter, that's how I like it :)

.env
```diff
@@ -0,0 +1,28 @@
+# In all environments, the following files are loaded if they exist,
```
Badlapje, line 1: should this file be committed?

Badlapje (commented):
Aside from one question, all is good as far as the eye can see.  Runs smoothly on my machine 👍 

Can't formally approve the PR as i'm the original author (which means github won't let me), but nice work & approval via text!

As far as the redesign goes, i've seen one small regression, which i'll solve in a separate PR ;-)

.env
MKodde: Allright, that is one of the things I need to address. Will fix this in a new PR, okay?

PR #175, opened 2021-06-29
Redesign profile

Badlapje:
Implement the redesign for profile

Prior to this change, profile looked old and gruffy.  Accessibility was low and UX hadn't received propper love in ages.

This change implements the redesign for profile accessibly.

Pivotal ticket @:


app/config/config_dev.yml
```diff
@@ -19,6 +19,7 @@ nelmio_security:
         enforce:
             img-src:
                 - self
+                - 'https://generative-placeholders.glitch.me'
```
pablothedude, line 22: Is this really needed?

app/web/css/components/modal.scss
```diff
@@ -0,0 +1,50 @@
+@import '../helpers/breakpoints';
+@import '../helpers/mixins';
+@import '../helpers/variables';
+
+// show the modal when the input is checked
+input[type="checkbox"]:checked.modal + label + .modal__value,
```
pablothedude, line 6: Be aware that by using + operators you're coupling the css to html. Also prevent the use of html elements (br). If you see fit to prevent that please do! This applies to all scss assets if applicable.

app/web/css/components/modal.scss
```diff
@@ -0,0 +1,50 @@
...
+
+@include ie11Only('input[type="checkbox"]:checked.modal + label + .modal__value.animated, input[type="checkbox"]:checked.modal + br + label + .modal__value.animated, input[type="checkbox"]:checked.modal + .button--tertiary + .modal__value.animated') {
+  max-height: 60vh;
+}
+
+label.modal {
+  cursor: pointer;
+}
+
+// set arrow down for modals when closed
```
pablothedude, line 39: Inline documentation :100:

app/web/js/app.js
```diff
@@ -1,9 +1,6 @@
 require('jquery');
-require('material-design-lite/dist/material.min.js');
-
+require('./dialog.js');
```
pablothedude, line 2: Not using Typescript? (same applies to all js assets below)

package.json
```diff
@@ -3,12 +3,13 @@
...
-    "jquery": "^3.6.0",
-    "material-design-lite": "1.0.5"
+    "jquery": "^3.6.0"
   },
   "devDependencies": {
     "@symfony/webpack-encore": "^1.5.0",
-    "file-loader": "^6.2.0"
+    "file-loader": "^6.2.0",
+    "sass": "^1.35.1",
+    "sass-loader": "^12.1.0"
```
pablothedude, line 12: Can you already get rid of the resolution below? package.json#20

src/OpenConext/ProfileBundle/Resources/views/MyConnections/active-connections.html.twig
```diff
@@ -0,0 +1,45 @@
+{% set activeConnections = {
```
pablothedude, line 1: Was this deliberate? If this was not for testing purposes  the activeConnections is empty check below could be removed else the set should be removed.

src/OpenConext/ProfileBundle/Resources/views/MyConnections/available-connections.html.twig
```diff
@@ -0,0 +1,39 @@
+{% set availableConnections = {
```
pablothedude, line 1: Was this deliberate? If this was not for testing purposes  the availableConnections is empty check below could be removed else the set should be removed.

pablothedude (changes requested):
Please address the feedback about the `set` in the twig templates. You could maybe add Typescript and appropriate linting and testing now because this seems the task to do it because you're now seem to touch everything. I leave that decision and the other feedback up to you. Nice work and :100: for introducing Sass!

app/config/config_dev.yml
Badlapje: definitely not, good catch!

app/web/css/components/modal.scss
Badlapje: Good one.  I actually did this for the tooltip, but forgot to remove it for the modal.  Greatly simplifies the css code as well :-)  💯

package.json
Badlapje: ah, i never thought to check, but it appears that yes i can ^__^

src/OpenConext/ProfileBundle/Resources/views/MyConnections/active-connections.html.twig
Badlapje: good catch x2!

app/web/js/app.js
Badlapje: I left JS for the next PR (or an addendum to this one).  Reasons for that:
- i'd like to strip out jQuery (given that it's just a bunch of simple clickhandlers really)
- i wanted to get some opinions on whether or not TS makes sense.  It feels to me that it's not worth the time, upkeep & added learning curve for people joining the project given the very limited amount of JS and the lack of any functions that actually would benefit from this.  There are two places in tooltip.js which could benefit, but those are properly guarded already.  Not sure about the old code, but in general i'd like to get rid of most of it.
- i'd like to get rid of the weird form for the language switcher and make those into plain links.  Not sure why it's a form now, so wanted to get your input on that & verify with the PM whether or not taking the time to do this makes sense.  From a semantic point of view the form makes no sense.  From a styling & code point of view i only see extra complications.  If i get rid of it, then that means most of the old JS is insta-gone.

pablothedude (approved):

MKodde (approved):

PR #180, opened 2021-07-28
Update translations for the redesign

Badlapje:
This change updates all translations for the redesign.  It also switches from an xliff format to a yml format.  The idea behind the switch is three-fold:
1. xliff is the format used by the web interface, but that was removed in 2019
2. yml is what we use in all other projects
3. yml is far easier to modify than the xliff files

A best effort was made for the Dutch translations.  For the Portugese translation, the keys that were changed are marked with "# FIXME" so the Portugese translator knows where to look.


pablothedude (approved):

Badlapje:
@domgon could you make the Portugese translations?

domgon:
Hello, I will translate the messages very soon

PR #181, opened 2021-07-30
translation to Portuguese

domgon:
the messages where translated to Portuguese and deleted the FIXME tag


Badlapje (approved):
Superb work, thx @domgon !

PR #182, opened 2021-08-17
Add delete modal for my connections

Badlapje:
Prior to this change, the old layout was still used.  The new design calls for a modal.

This change ensures the modal with the propper design is used.  It also ensures the modal is fully accessible.

The modal technique was taken from this example: https://tutorialspage.com/create-modal-window-pure-html5-css3/
The modal trap was modified from code found on: https://uxdesign.cc/how-to-trap-focus-inside-modal-to-make-it-ada-compliant-6a50f9a70700

### DIFF
--- a/app/web/css/pages/myConnections/myConnections.scss
+++ b/app/web/css/pages/myConnections/myConnections.scss
@@ -7,3 +7,8 @@
 .connection__buttonWrapper {
   margin-top: 2rem;
 }
+
+.connection__deleteForm {
+  margin-left: 1.5rem;
+  width: calc((100% - 1.5rem) / 2);
+}
--- a/app/web/js/my-connections.js
+++ b/app/web/js/my-connections.js
@@ -1,11 +1,11 @@
-$(document).on('click', 'button.disconnect', function (e) {
-    var button = $(e.target),
-        form = $('form.confirmation-container');
-
-    button.hide(200, function(){
-        form.show(200);
-    });
-});
+// $(document).on('click', '.disconnect', function (e) {
+//     var button = $(e.target),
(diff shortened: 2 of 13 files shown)


app/web/css/pages/myConnections/myConnections.scss
```diff
@@ -7,3 +7,8 @@
 .connection__buttonWrapper {
   margin-top: 2rem;
 }
+
+.connection__deleteForm {
+  margin-left: 1.5rem;
+  width: calc((100% - 1.5rem) / 2);
```
pablothedude, line 13: Does the line above work in all supported browsers, I saw it multiple times also in other files? Just wondering... Feel free to leave it as-is.

app/web/js/my-connections.js
```diff
@@ -1,11 +1,11 @@
-$(document).on('click', 'button.disconnect', function (e) {
-    var button = $(e.target),
-        form = $('form.confirmation-container');
-
-    button.hide(200, function(){
-        form.show(200);
-    });
-});
+// $(document).on('click', '.disconnect', function (e) {
```
pablothedude, line 1: If this is death code, please remove.

pablothedude (approved):
:+1: Nice work!

app/web/css/pages/myConnections/myConnections.scss
Badlapje: This line does indeed, i have tested it in IE11 (which is the only one with partial support of the calc function), works like a charm.  Not so for the entire redesign, i'll have to make some commits to fix the errors on other parts of the page.

PR #186, opened 2021-08-27
Ensure cross-browser functionality & styling

Badlapje:
Prior to this change, cross-browser wasn't tested adequately.

This change ensures cross-browser compatibility:
- ensure the cards on the home page look good in IE 10/11
- ensure the download button looks good in Safari
- ensure the download button is left-aligned on mobile
- ensure the idpRow focus style is good in Safari
- ensure the modals look good on mobile


MKodde:
Hi @Badlapje, nice work again. 

I tried testing in BrowserStack but this turned out to be a bridge to far. My local network was unreachable in the BrowserStack browsers. So I have not scrutinized your changes on any other relevant browsers.

**Some remarks I was able to sniff out on my machine.** 
1. Pressing the turned 90 degrees hamburger menu chevron thingy does not actually expands the menu. But clicking the blue ribbon beneath it does. [Firefox on Linux in Mobile simulation mode]
2. Sometimes after opening/closing the menu a couple of times the icons remain visible. See screenshot. [Firefox on Linux in Mobile simulation mode]

Screenshot from 2021-08-31 14-44-10

Badlapje:
I found out about the first issue when working on the accessibility changes.  So i fixed it there (PR forthcoming).

As for problem 2: i can't replicate this behaviour on either firefox/chrome/edge when simulating it.  I'd await this as a possible finding from testing on test2 if that's ok with you?

MKodde:
| As for problem 2: i can't replicate this behaviour on either firefox/chrome/edge when simulating it. I'd await this as a possible finding from testing on test2 if that's ok with you?

Sure thing! Hopefully this was just a browser cache issue on my end of the line.

:ship: :clown_face: = :shipit:

MKodde (approved):

PR #187, opened 2021-08-30
Update lockfile to version 2

MKodde:
This because I started using a newer node version. And npm started
nagging about thhe old lockfileversion. For me this is no issue, but
might restrict older users from building a working set of packages?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,8284 @@
 {
   "name": "openconext-profile",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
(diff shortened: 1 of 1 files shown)


MKodde:
Is this what we want? And should we update NPM version restrictions in the NPM `package.json` and the readme?

MKodde:
After reading up on NPM versions and the associated lock file versions. I came to the conclusion it best to close this PR and await the moment we want to migrate to NPM 7 to also update the lockfiles.

TIL: 
NPM < v7 generates lockfileversion 1
NPM >= v7 generates lockfileversion 2 or 3. Where lockfileversion 2 is BC with older NPM versions and version 3 is not.

PR #189, opened 2021-09-02
Fix accessibility problems

Badlapje:
Prior to this change, accessibility was good, but not thoroughly tested.

This change adds automatic tests using cypress, axe and html-validator & adds those to the CI.
It also fixes all problems found in manual & automatic testing:
- mobile menu didn't work always on tap
- pressing the space bar didn't activate the details element
- something incorrect modal didn't work with the space bar
- some colour contrast issues
- adds focusable to the arrow svg (mandatory attribute)
- add text to the logo-link
- add type to the language-switcher buttons
- add type to the mobile navigation button
- ensure the mobile menu isn't hidden when no js is available
- add a label to the language switcher

To earn a boyscout badge, a problem with the language switcher was also fixed (where it didn't work anymore).

### DIFF
--- a/assets/css/components/locale-switch.scss
+++ b/assets/css/components/locale-switch.scss
@@ -6,9 +6,14 @@
   width: 33%;
 }
 
+label.locale-switch-label,
 select[name="profile_switch_locale[newLocale]"],
 button[type="submit"][name="profile_switch_locale[changeLocale]"] {
   display: none;
+}
+
+.no-js label.locale-switch-label {
+  display: inline-block;
 }
 
 .no-js select[name="profile_switch_locale[newLocale]"],
--- a/assets/css/components/note.scss
+++ b/assets/css/components/note.scss
@@ -10,3 +10,7 @@
   margin: 0;
   padding: 1rem;
 }
+
+.note__link {
+  color: $darkishBlue;
+}
--- a/assets/css/helpers/classes.scss
+++ b/assets/css/helpers/classes.scss
@@ -11,11 +11,12 @@
 }
 
 .skip-main {
-    @include absolute(top 0 right 0);
-    color: $oceanBlue;
+    @include visually-hidden;
 
     &:focus,
     &:active {
+        @include visually-shown;
+        @include absolute(top 0 right 0);
         color: $white;
         margin: 10px;
     }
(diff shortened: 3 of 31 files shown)


app/Resources/translations/messages.nl.yml
```diff
@@ -193,7 +194,7 @@ profile:
 
     navigation:
         help: Help
-        privacy: profile.navigation.privacy
+        privacy: Privacy
```
MKodde, line 197: :+1: for  getting rid of an untranslated translation

app/web/js/localeSwitch.js
```diff
@@ -41,7 +56,7 @@ window.addEventListener('load', function () {
                             const optionElement = optionElements[k];
                             if (optionElement.value === choice) {
                                 optionElement.setAttribute('selected', 'selected');
-                                return true;
+                                continue;
```
MKodde, line 59: Is this change in behavior intended; Breaking out of the function VS continuing it's routine? 

The way I see it, it now adds the selected class and then removes it right afterwards. But I have not studied this code in super great detail. Just noticing a little smell. Please elaborate.

MKodde (changes requested):
Your changes are looking great. 

Aside from my question below, I think you should remove the example Cypress tests from the project. Or rename them if they have any value to us.

Badlapje:
> 
> 
> Your changes are looking great.
> 
> Aside from my question below, I think you should remove the example Cypress tests from the project. Or rename them if they have any value to us.

done

app/web/js/localeSwitch.js
Badlapje: It's intentional.  

What it does is one of two things:
- it sets the selected attribute if the value of the element = the choice of the user.  In that case the continue ensures that only the if statement is executed in that iteration of the for loop (which means the selected attribute isn't removed again just after setting it.
- it removes the selected attribute in case the value does not correspond with the choice of the user.  This is necessary (and is what didn't happen before), because one of the option-elements has the selected class given that it corresponds to the current value, rather than the value the user desires.

MKodde: Allright no objections from my side then! Thanks for making things clear.

MKodde (approved):

PR #191, opened 2021-09-06
Deprecation resolutions

MKodde:
Remove as many deprecation warnings as humanly possible. A large amount of warnings are left in place as they are generated by the Stepup-bundle


Badlapje (approved):
:shipit: Gotta love less deprecation resolutions 😄 And it seems to work just fine still ^__^  Nice work!

PR #192, opened 2021-09-06
Ensure login details is on one row

Badlapje:
Prior to this change, there was a regression in the styling of login-details.  The label received too little space, meaning it was split over two rows.

This change ensures the label is fully on one row again.

### DIFF
--- a/assets/css/pages/myServices/loginDetails.scss
+++ b/assets/css/pages/myServices/loginDetails.scss
@@ -10,7 +10,7 @@
     color: $black;
     font-weight: $normal;
     line-height: calculateRem(31px);
-    width: 25%;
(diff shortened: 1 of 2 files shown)


MKodde:
For clarification, the fix was issued on the section seen in the screenshot below.

image (2)

MKodde (approved):
Sweet and simple!

No issues with merging this. Related to this issue I found a small nit for the mobile viewport.

Screen Shot 2021-09-09 at 10 29 42

On mobile, the value column has excessive padding to the left. Making it jump out of the gutter. To be clear: I'm talking about the values of the `First used on` and `Consent was given by` data fields.

PR #193, opened 2021-09-08
Fix problems in IE11

Badlapje:
While testing the design for regressions resulting from the upgrade to SF4, it was noticed that the IE11 meta tag was missing.  After adding that tag, the design was tested thoroughly and several extra problems were found.

This change resolves all found issues:
- added the ie meta tag
- used insertBefore instead of prepend to add a label for the select, as prepend is not supported in IE
- added a polyfill for the details element
- rectified styling issues for the details element
- fix styling of download button in IE11
- ensure icons are visible in IE11
- optimize svg code as a result of the above
- ensure surf logo is visible in header and footer for IE11
- optimize width of attributes in table of my services
- fixed view of "something incorrect" for IE11

See https://github.com/javan/details-element-polyfill for the polyfill used


MKodde (approved):
The code changes are looking great! I've clicked through the application using an IE11 on Windows 10. I found four possible things you might want to look into. But might be insignificant.. You and PM decide what's good enough..

ProfileIE11-1

1. Icons are missing in the menu also after clearing the cache and re-running encore several times.

ProfileIE11-2

2. Column is not wide enough, or additional margin/padding should be applied to the label?
3. Think this was fixed in #192.. But might need some additional ie11 TLC?

ProfileIE11-3

4. Something incorrect is mangled up.

Badlapje:
1 don't quite see how i missed this one 🤯 But fixed it after some searching, along with the logo of surf both in header & footer.  While i was at it, i optimized some of the svg code to make the file-size smaller.
2 was fixed with a minimal procedure
3 is fixed by #192 
nice catch on 4: fixed that.

PR #194, opened 2021-09-08
Add back missing images

Badlapje:
Prior to this change, the upgrade to symfony 4 screwed up some images.

This change ensures all missing images are visible again.


MKodde (approved):
I'm guilty as charged. Thanks for dragging the imagery back to the assets folder!

PR #196, opened 2021-09-09
Feature/GitHub actions build docker

quartje:
This PR will add two GitHub action workflows: One to build and publish the httpd and php fpm repository, and one to clean up old packages from the container registry.

Packages are built once a PR is opened, when a push for the branch "develop" is executed, and when a version is tagged

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -1,0 +1,18 @@
+name: Remove old package versions
+on:
+  workflow_dispatch:
+    inputs:
+      package_name:
+        description: 'The name of the package to delete'     
+        required: true
+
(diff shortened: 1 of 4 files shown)


MKodde (approved):
Thanks for this effort @quartje looking very promising!

Would this PR be a good place to update and test the `docker/Dockerfiledev` file with the new Profile container?

quartje:
These are different containers: The production containers do have all application code in it, something that is not in the dev container (where the application files would typically be on your own machine). I do have another project running, OpenConext-containers, where I plan to build containers that can form the foundation of other containers: This would eliminate the need for building the dev containers, you'd just run them.

PR #197, opened 2021-09-13
Setup GitHub Actions for CI

MKodde:
Using Github Actions in favor of Travis makes sense on several levels. For one we can easily use the docker containers already available to us. Making the development and CI environment very similar. Yielding more predicable test results from the CI env. 

This PR aims to move the test coverage we enjoy on Travis to GH Actions.


Badlapje (approved):
Looks great to me :-)  Nice work!

PR #198, opened 2021-09-13
Improve cypress tests

Badlapje:
Prior to this change there wasn't much logging when cypress ran in headless mode.  We also had some rules in html-validate which threw incorrect errors.  As a result those were configured to only throw warnings. Thankfully, those errors were solved by the project maintainer of html-validate.

This change:
- adds a plugin to increase logging in headless mode.
- upgrades html-validate en cypress-html-validate to the new & improved version.
- ensures the warnings are errors again and adds a specific rule for heading level.
- fixes an error found when testing for the above fix: it turns out there was no document language set on the html tag.  Now there is.

### DIFF
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -1,24 +1,20 @@
 const htmlvalidate = require('cypress-html-validate/dist/plugin');
 
 module.exports = (on, config) => {
+    require('cypress-terminal-report/src/installLogsPrinter')(on);
+
     htmlvalidate.install(on, {
         "rules": {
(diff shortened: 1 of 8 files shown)


MKodde (approved):
I recon this is a PR on top of another PR? The locale changes seem unrelated to the cypress-terminla-report installation.

Aside from that, all looks grand! :shipit:

MKodde:
@Badlapje Before merging this, can we investigate how we can run the Cypress a11y tests on Github Actions (now part of your branch after a `git reset`)

PR #199, opened 2021-09-14
Docker: Use OpenConext containers to build the production containers

quartje:
This will use the Docker containers which are built in the OpenConext-container project.


MKodde (approved):
Superfly!

PR #203, opened 2021-10-04
Update release script and general cleanup

MKodde:
The release script was updated, references to Symfony 3 files where removed. And new Symfony 4 concepts are covered/added in the PR.

Some remarks and additional tasks
- Ansible and Vagrant files are removed in favor of docker files
- Readme was updated (vagrant instructions have been removed)
- Previously we would run the release script in the vagrant box. Is a release from the developers bare metal also safe enough? Or can we lever our Docker container to build a release?
- This PR must be rebased onto `feature/redesign` once we reach maturity there.


quartje:
Nice work, it cleans up nicely :)
For the building part: Some commands have changed, so I recommend we use the same as is done in Docker/dockerfile

Considering the build on Vagrant: If we move the makeRelease script to it's own directory (say Releases), a build can be triggered using this simple oneliner (here is an example building the develop branch):

docker run  -v  $PWD/Releases/:/root/Releases/ ghcr.io/openconext/openconext-containers/openconext-phpfpm-dev:latest  /root/Releases/makeRelease.sh develop

The tarball will be in your local Releases folder.

MKodde:
Thanks @quartje, I added the instructions to the readme!

quartje (approved):
Right, that seems to fix the build script. As discussed, some feature enhancements:

PR #204, opened 2021-10-06
Makerelease: Move to seperate folder, and use in stead of bash

quartje:
Two small fixes:
* Move makeRelease.sh to  a seperate folder that can be mounted into the docker build container
* Use sh in stead of bash, the latter is not available in the container


Badlapje (approved):
ziet er goed uit: mooi werk!

PR #221, opened 2021-10-27
Allow overriding translations

Badlapje:
Prior to this change, there was no way to override suite_name and organisation_noun translations.

This change moves towards the inbuilt symfony translations in a php format.  This allows us to override the translations via the same mechanism as used in engineblock.

Pivotal ticket @: https://www.pivotaltracker.com/story/show/180095734


MKodde:
I love this change, it syncs the EngineBlock methodology with the one used for Profile. 

This is a rather big change, and can be considered BC breaking maybe? At this moment I find it hard to wrap my brain around this. As translations can no longer be provided with the YAML format. And extraction of translations are no longer supported.

1. Is this intentional? And the best option?
2. Was OpenConext-deploy also updated? The translations used in the templates there might not be converted from YAML to PHP. 
3. How would a potential rollback scenario look like?

thijskh:
For a 3.0 release I am OK with breaking BC if we move in one go to a solution that is already present and known to work in EB.

I don't think OpenConext-deploy contains profile translations, probably precisely because it was not possible to override them in Profile <3

PR #222, opened 2021-11-04
Use correct institution data

Badlapje:
Prior to this change the institution was always shown as "Ibuildings university" with the surf logo.

This change ensures we retrieve the correct data from the engineblock api and show it in the my services view.

Pivotal ticket @ https://www.pivotaltracker.com/story/show/180095716

### DIFF
--- a/ci/npm-audit.sh
+++ b/ci/npm-audit.sh
@@ -1,2 +1,2 @@
 #!/bin/sh
-npm audit
+npm audit --production
--- a/src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
+++ b/src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
@@ -25,6 +25,7 @@ use OpenConext\EngineBlockApiClientBundle\Http\JsonApiClient;
 use OpenConext\Profile\Entity\AuthenticatedUser;
 use OpenConext\Profile\Value\EntityId;
 use OpenConext\Profile\Value\Logo;
+use OpenConext\Profile\Value\Organization;
 use Psr\Log\LoggerInterface;
 
 final class InstitutionRepository
@@ -65,14 +66,17 @@ final class InstitutionRepository
         }
     }
 
-    public function getDisplayNameAndLogoForIdp(AuthenticatedUser $user, string $locale): array
+    /**
+     * @throws AssertionFailedException
+     */
+    public function getOrganizationAndLogoForIdp(AuthenticatedUser $user): array
     {
         $entityIds = $user->getAuthenticatingAuthorities();
         $authenticatingIdpEntityId = $this->getNearestAuthenticatingAuthorityEntityId($entityIds);
         $json = $this->findAllForIdp($authenticatingIdpEntityId->getEntityId());
 
         return [
-            'displayName' => $this->getDisplayName($json, $locale),
+            'organization' => $this->getOrganization($json),
             'logo' => $this->getLogo($json),
             'json' => $json,
         ];
@@ -111,23 +115,8 @@ final class InstitutionRepository
         return false;
     }
 
-    private function getDisplayName(array $json, string $locale): string
+    private function getOrganization(array $json): Organization
     {
-        $displayNameLocale = $json['display_name'][$locale];
-        if (!empty($displayNameLocale)) {
-            return $displayNameLocale;
-        }
-
-        $localeName = $json['name'][$locale];
-        if (!empty($localeName)) {
-            return $localeName;
-        }
-
-        $englishName = $json['name']['en'];
-        if (!empty($englishName)) {
-            return $englishName;
-        }
-
-        return '';
+        return Organization::fromArray($json);
     }
 }
--- a/src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
+++ b/src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
@@ -43,7 +43,7 @@ final class InstitutionRepository
         $this->engineBlockEntityId = $engineBlockEntityId;
     }
 
-    public function findAllForIdp(string $entityId)
+    private function findAllForIdp(string $entityId)
     {
         try {
             return $this->apiClient->read('metadata/idp?entity-id=%s', [$entityId]);
--- a/src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
+++ b/src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
@@ -18,11 +18,13 @@
 
 namespace OpenConext\EngineBlockApiClientBundle\Repository;
 
+use Assert\AssertionFailedException;
 use Exception;
 use OpenConext\EngineBlockApiClientBundle\Exception\ResourceNotFoundException;
 use OpenConext\EngineBlockApiClientBundle\Http\JsonApiClient;
 use OpenConext\Profile\Entity\AuthenticatedUser;
 use OpenConext\Profile\Value\EntityId;
+use OpenConext\Profile\Value\Logo;
 use Psr\Log\LoggerInterface;
 
 final class InstitutionRepository
@@ -72,6 +74,7 @@ final class InstitutionRepository
         return [
             'displayName' => $this->getDisplayName($json, $locale),
             'logo' => $this->getLogo($json),
+            'json' => $json,
         ];
     }
 
@@ -95,13 +98,14 @@ final class InstitutionRepository
     }
 
     /**
-     * @return false|array
+     * @return false|Logo
+     * @throws AssertionFailedException
      */
     private function getLogo(array $json)
     {
         $logo = $json['logo'];
         if (!empty($logo['url'])) {
-            return $logo;
+            return Logo::fromArray($logo);
         }
 
         return false;
--- a/src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
+++ b/src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
@@ -19,9 +19,11 @@
 namespace OpenConext\EngineBlockApiClientBundle\Repository;
 
 use Exception;
+use OpenConext\EngineBlockApiClientBundle\Exception\ResourceNotFoundException;
 use OpenConext\EngineBlockApiClientBundle\Http\JsonApiClient;
 use OpenConext\Profile\Entity\AuthenticatedUser;
 use OpenConext\Profile\Value\EntityId;
+use Psr\Log\LoggerInterface;
 
 final class InstitutionRepository
 {
@@ -35,12 +37,17 @@ final class InstitutionRepository
      */
     private $engineBlockEntityId;
 
+    /** @var LoggerInterface */
+    private $logger;
+
     public function __construct(
         JsonApiClient $apiClient,
-        EntityId $engineBlockEntityId
+        EntityId $engineBlockEntityId,
+        LoggerInterface $logger
     ) {
         $this->apiClient = $apiClient;
         $this->engineBlockEntityId = $engineBlockEntityId;
+        $this->logger = $logger;
     }
 
     private function findAllForIdp(string $entityId)
@@ -48,7 +55,11 @@ final class InstitutionRepository
         try {
             return $this->apiClient->read('metadata/idp?entity-id=%s', [$entityId]);
         } catch (Exception $e) {
-            return $e->getMessage();
+            $this->logger->notice(
+                sprintf('EngineBlock API returned a non 200 response with error message (%s)', $e->getMessage())
+            );
+
+            throw new ResourceNotFoundException();
         }
     }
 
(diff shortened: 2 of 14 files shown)


src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
```diff
@@ -0,0 +1,118 @@
...
+            return $localeName;
+        }
+
+        $englishName = $json['name']['en'];
+        if (!empty($englishName)) {
+            return $englishName;
+        }
+
+        return '';
+    }
```
MKodde, line 117: Think this logic should be included in the DisplayName value object. Or alternatively if that Value Object is not practical in use, create one that fits your needs and include this logic there.

src/OpenConext/ProfileBundle/Controller/MyServicesController.php
```diff
@@ -88,17 +100,25 @@ public function overviewAction(Request $request)
 
         $this->logger->notice('User requested My Services page');
 
+        $locale = $request->getLocale();
         $user = $this->authenticatedUserProvider->getCurrentUser();
         $specifiedConsentList = $this->specifiedConsentListService->getListFor($user);
-        $specifiedConsentList->sortByDisplayName($request->getLocale());
+        $specifiedConsentList->sortByDisplayName($locale);
```
MKodde, line 106: :ok_hand:

src/OpenConext/ProfileBundle/Controller/MyServicesController.php
```diff
@@ -88,17 +100,25 @@ public function overviewAction(Request $request)
...
 
         $this->logger->notice(sprintf('Showing %s services on My Services page', count($specifiedConsentList)));
 
+        list(
+            'displayName' => $displayName,
+            'logo' => $logo
+        ) = $this->institutionRepository->getDisplayNameAndLogoForIdp($user, $locale);
```
MKodde, line 113: Time to roll in the VO parade. Making these wonky constructions a thing of the past :elephant:

src/OpenConext/ProfileBundle/Resources/config/services.yml
```diff
@@ -17,6 +17,8 @@ services:
         arguments:
             - ~ # engine block entity id
 
+    OpenConext\EngineBlockApiClientBundle\Repository\InstitutionRepository:
```
MKodde, line 20: Enabling autoconfigure or autowire, I'm uncertain which one does what exactly, on the default container should mitigate having to do this.

MKodde (commented):
You did a great job partner :cowboy_hat_face: 

I do feel you should let the new repo return value objects instead of way to raw json decoded array. Other than that this is good to go.

src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
```diff
@@ -0,0 +1,118 @@
...
+
+    public function __construct(
+        JsonApiClient $apiClient,
+        EntityId $engineBlockEntityId
+    ) {
+        $this->apiClient = $apiClient;
+        $this->engineBlockEntityId = $engineBlockEntityId;
+    }
+
+    public function findAllForIdp(string $entityId)
```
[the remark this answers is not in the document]
Badlapje: good catch 👍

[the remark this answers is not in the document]
Badlapje: fixed i think

src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
Badlapje: done

src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
```diff
@@ -0,0 +1,118 @@
...
+    public function getDisplayNameAndLogoForIdp(AuthenticatedUser $user, string $locale): array
+    {
+        $entityIds = $user->getAuthenticatingAuthorities();
+        $authenticatingIdpEntityId = $this->getNearestAuthenticatingAuthorityEntityId($entityIds);
+        $json = $this->findAllForIdp($authenticatingIdpEntityId->getEntityId());
+
+        return [
+            'displayName' => $this->getDisplayName($json, $locale),
+            'logo' => $this->getLogo($json),
+        ];
```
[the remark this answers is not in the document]
Badlapje: fixed

src/OpenConext/ProfileBundle/Resources/config/services.yml
Badlapje: right again, it's autowire

Badlapje: weird enough when i first tested this it worked, but after pushing it and trying it a second time it fails every time.  So i kept it in there.  Autowiring aparently doesn't work for this class, not sure why.

MKodde: No hard feelings! On the next major release of Symfony I believe we need to overhaul the container configuration. So we can fix it there and then.

ci/npm-audit.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-npm audit
+npm audit --production
```
MKodde, line 2: :heart: thats point one from the stand-up todo list done and dusted

src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
```diff
@@ -0,0 +1,94 @@
...
+    private function findAllForIdp(string $entityId)
+    {
+        try {
+            return $this->apiClient->read('metadata/idp?entity-id=%s', [$entityId]);
+        } catch (Exception $e) {
+            $this->logger->notice(
+                sprintf('EngineBlock API returned a non 200 response with error message (%s)', $e->getMessage())
+            );
+
+            throw new ResourceNotFoundException();
```
MKodde, line 63: To be more efficient. You can remove the log entry and set the text on the exception message. Exceptions are also logged so the entry will end up in logs that way. And you also lose the dependency on the loggerinterface that way.

src/OpenConext/Profile/Value/Logo.php
```diff
@@ -0,0 +1,79 @@
...
+     * @var string
+     */
+    private $height;
+
+    /**
+     * @var string
+     */
+    private $width;
+
+    public function __construct(?string $url, ?string $height, ?string $width)
```
MKodde, line 41: Are we required to allow null values for these params? In the static constructor we force them to be present. So maybe you can remove the `?`?

src/OpenConext/Profile/Value/Logo.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public function hasUrl(): bool
+    {
+        return !empty($this->getUrl());
+    }
+
+    /**
+     * @return string
+     */
```
MKodde, line 66: Feel free to remove the phpdoc as the return type definition will help our IDE's in recognizing the return type

src/OpenConext/Profile/Value/Organization.php
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * @return false|Logo
+     */
+    public function getLogo()
+    {
+        if (!$this->hasLogo()) {
+            return false;
+        }
```
MKodde, line 86: Having a mixed return type is not encouraged. I'd simply let the method throw an exception when the logo is not set. Which should never be the case, as you can test this using the 'hasLogo' method.

src/OpenConext/ProfileBundle/Resources/views/MyServices/AttributeList/idp.html.twig
```diff
@@ -58,14 +58,14 @@
...
 <li class="idpRow">
-    {% set logoUrl = ('provided_by_attribute_source_logo_url_' ~ sourceName)|trans %}
-    {% set organisationName = ('provided_by_attribute_source_' ~ sourceName)|trans %}
     {% set consentProvidedBy %}
-        <span class="idpRow__providedBy-content">{{ 'profile.my_services.idpRow.consent_provided_by'|trans }} <strong>{{ organisationName }}</strong></span>
+        <span class="idpRow__providedBy-content">{{ 'profile.my_services.idpRow.consent_provided_by'|trans }} <strong>{{ displayName }}</strong></span>
     {% endset %}
 
-    <img class="logo-small" src="{{ asset(logoUrl) }}"
+    {% if logo %}
```
MKodde, line 65: I'd say use: `logo.hasLogo` here instead. See my other comment on this in the Value object.

MKodde (changes requested):
Nearly there @Badlapje. There are some things that need some work. Also, not mentioned in my code review. I expect you to add some unit test coverage for the new value objects. They are easily tested and will allow you to fix some of my feedback in a controlled manner.

src/OpenConext/Profile/Value/Logo.php
Badlapje: We are, as the feedback from the engineblock api can contain null values for these params.

src/OpenConext/Profile/Value/Logo.php
Badlapje: done

src/OpenConext/ProfileBundle/Resources/views/MyServices/AttributeList/idp.html.twig
Badlapje: fixed, i like it 👍

src/OpenConext/Profile/Value/Organization.php
Badlapje: seems so obvious in hindsight, good call 💯

src/OpenConext/EngineBlockApiClientBundle/Repository/InstitutionRepository.php
Badlapje: done :-)  Didn't know that exceptions get logged, good to know!

src/OpenConext/Profile/Tests/Value/LogoTest.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * @test
+     * @group Logo
+     * @group Value
+     */
+    public function hasUrlReturnsFalseWhenThereIsNoUrl()
+    {
+        $logoWithoutUrl = new Logo(null, null, null);
+        $otherLogoWithoutUrl = new Logo('', '', '');
```
MKodde, line 34: Hurray for adding the test mate! 

I still think using the Logo value object this way is fundamentally wrong. I'd rather see an EmptyLogo and a Logo variety. Whats the point of a Logo value object that does not represent a logo? 

Still feel free to leave this change the way it is if refactoring costs extra (too much) time. This solution works, and is easy enough to understand.

src/OpenConext/Profile/Tests/Value/LogoTest.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * @test
+     * @group Logo
+     * @group Value
+     */
+    public function hasUrlReturnsTrueWhenUrlIsPresent()
+    {
+        $logoWithUrl = new Logo("https//fake.logo/logo.jpg", null, null);
```
MKodde, line 46: As far as I know, we do not use the width and height properties of the logo. Maybe Its better to scratch them altogether. Unused code serves no purpose. If ever we need them they can be added at that point without issue

Again, not a blocker, just sayin'

src/OpenConext/Profile/Tests/Value/OrganizationTest.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->assertEquals($organization->getDisplayName('pt'), 'English name');
+        $this->assertEquals($organization->getDisplayName('nl'), 'Nederlandse displaynaam');
+
+        $displayName['nl'] = '';
+        $organization2 = new Organization($displayName, $name, $logo);
+        $this->assertEquals($organization2->getDisplayName('nl'), 'Nederlandse naam');
+
+        $name['nl'] = '';
+        $organization3 = new Organization($displayName, $name, $logo);
+        $this->assertEquals($organization3->getDisplayName('nl'), 'English name');
```
MKodde, line 49: To my knowledge it was not mentioned in the story that when a specific languages name or display name is not present we should fall back to the English 'name:en' value. I'd say first try the English display name and then the 'regular' English name. But this is fine too I guess.

Badlapje: https://www.pivotaltracker.com/story/show/180095716 says:

>  "displayname:$lang"
> "name:$lang"
> "name:en"
> static string "Your %organizationNoun%"
> 
> In other words, if current lang is PT, see if we have a PT displayname, otherwise a PT name. If not, use the English name (in theory always set/mandatory). For final robustness return static string if none found.

src/OpenConext/Profile/Tests/Value/LogoTest.php
Badlapje: adjusted

MKodde (approved):

PR #223, opened 2021-11-23
Downgrade 'normal operation' log messages to info

thijskh:
These are the kinds of situations that are just run of the mill (show a specific
page, mostly) and do not need to be logged above the passthrough level. In the
development environment one can override the passthrouh level to show these
if necessary.


Badlapje (approved):
Looks good, nice work!

PR #224, opened 2021-11-24
Fix IdP logo display in My Services overview

thijskh:
hasLogo() does not exist so is always false, hasUrl() does exist.


Badlapje (approved):
💯

PR #225, opened 2022-01-17
Create release artifacts when project is tagged

MKodde:
In this workflow an incoming tag is triggering the makeRelease script. The output of this script (the artifacts) are used to create a new release page on the projects GitHub releases page. The artifacts are then uploaded to that new release.

The one thing remaining is to add the release notes to the release. This is now set with a small placeholder message

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -1,0 +1,55 @@
+name: tag-release
+on:
+  push:
+    tags:
+      - "*.*.*"
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    if: always()
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: Output the semver tag to the tag variable
+        id: vars
+        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+      - name: Pull OpenConext build container
+        run: docker-compose -f ./docker-compose-tag-release.yml up -d
+      - name: Make the release files
+        run: docker-compose exec -T openconext sh -c 'HOME=/var/www/html ./bin/makeRelease.sh ${{ steps.vars.outputs.tag }}'
+      - name: Create Draft Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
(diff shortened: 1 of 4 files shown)


quartje (commented):
Nice work!
I see that you first use "draft" and then a regular release in the last step. Is that by design?
Did you test the resulting tar.gz to check if it contains what you want? 

I think that in a future change we might take a look into makeRelease.sh to make it explicit what needs to be _in_ the tarball, in stead of the other way around and delete what should not be in it.

.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,65 @@
...
+        id: vars
+        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+      - name: Check output
+        env:
+          RELEASE_VERSION: ${{ steps.vars.outputs.tag }}
+        run: |
+          echo $RELEASE_VERSION
+          echo ${{ steps.vars.outputs.tag }}
+      - name: Init environment
+        run: cd docker && ./init.sh && cat .env && cd ../
```
quartje, line 24: Is this step necessary?

MKodde: Good point! That was not mandatory. Also, I removed the 'List items' step. That was also a debug left-over

pablothedude:
Hi @MKodde, I think a great addition, it certainly will reduce the hassle when releasing and will cost a lot less effort.

The only thing you possibly should take into account is that currently the tarring works by removing some files/folders before tarring. But consider to do the opposite by only adding what you want in the tar. This in order to prevent the possibility that files are accidentality added during a release.

But all in all great work!

docker-compose-tag-release.yml
```diff
@@ -0,0 +1,12 @@
+---
+
+version: "3.8"
+
+services:
+  openconext:
+    image: "ghcr.io/openconext/openconext-containers/openconext-phpfpm-dev:latest"
```
pablothedude, line 7: Nice to see this!

pablothedude (approved):

MKodde:
Thanks for your feedback @pablothedude Bart had the same concern, and we are planning to do just that in the near future.

PR #227, opened 2022-01-20
Add the option to store sessions in the database

quartje:
When sticky sessions are not an option it might be usefull to store sessions in the database. This adds a configuration parameter to do so.

The default is still file based session storage


thijskh:
Would it be an idea to just make the database dsn the thing you configure instead of the broken down parts? Then you can instantly influence all other aspects from config, even the backend db type (as I know there are OpenConext installs using non-MySQL backends)

MKodde (approved):
LGTM, I must say I like Thijs' idea. That might simplify things slightly.

quartje:
I have changed it, now you can configure the whole dsn

PR #233, opened 2022-04-13
Remove SURFisms from translation files

MKodde:
Replacing the SURFisms with a configurable organisation noun and suite name was already prepared in a previous PR by @Badlapje. The work was not finalized. As the translations where not yet applied on the translation texts. This PR takes care of that. Utilizing the i18n Twig extension. See commit messages and the story for additional details.

:warning: One point of attention @thijskh the `introduction.explanation.introduction.end` translation was translated a bit fishy.

`'you login with your institutional account on different (cloud)services. You can find more info here.',`
After
`'you login with your %organisationNoun%al account on different (cloud)services. You can find more info here.',`

This works out with both organisation nouns: institution and organisation. But use of other org-nouns might result in a grammatical error.

Should we re-write the sentence?

See: https://www.pivotaltracker.com/story/show/180269047


thijskh:
Let's accept that for now since I think the chances of adding more nouns in the forseeable future are quite slim.

translations/messages.nl.php
```diff
@@ -24,18 +24,18 @@
         'cancel' => 'Annuleren',
         'confirm' => 'Bevestigen',
         'home' => 'Naar de homepagina',
-        'organisation_noun' => 'organisatie',
-        'suite_name' => 'OpenConext',
+        'organisation_noun' => 'bibliotheek',
+        'suite_name' => 'BiebConext',
```
thijskh, line 28: Should this be actually committed?

MKodde: Lolz, that was my test setup. Will revert that! Thanks for spotting that

VadimSchmitz (approved):

PR #234, opened 2022-04-14
Prevent unsafe-inline CSP errors on SVG images

MKodde:
SVG images that are included on the page would trigger CSP usafe inline errors. Loosening the CSP would open us to a host of other vulnerabilities. So fixing the SVG's was the way to go.

Having the inline styles in a dedicated stylesheet proved to be the most pragmatic solution. Thanks @Badlapje for thinking out of the box with me on this one! 

Bugreport: https://www.pivotaltracker.com/story/show/180407180

### DIFF
--- a/assets/css/pages/myServices/serviceList.scss
+++ b/assets/css/pages/myServices/serviceList.scss
@@ -41,9 +41,3 @@
   }
 }
 
-.support__icon {
-  fill: none;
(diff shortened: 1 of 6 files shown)


Badlapje (approved):
Looks good to me 👍

thijskh (approved):

PR #243, opened 2022-05-09
Make the cache path `var/cache/prod` again instead of `var/prod/cache`.

thijskh:
The swithed order was introduced in 3.0 but for no apparent reason, since var/cache/<env> is still the default used everywhere, this was likely by mistake (maybe from copying [this example](https://symfony.com/doc/current/configuration/override_dir_structure.html#override-the-cache-directory)?) so change it back so it matches what is customary.


Badlapje (approved):
Looks good to me :shipit:

MKodde (approved):
This indeed makes more sense! Thanks for changing this back to normal :+1:

PR #284, opened 2024-04-25
Add JsonFormatter

parijke:
Add a JsonFormatter to the project and use it in monolog


MKodde (commented):
This seems to fit the bill. 

Question: why not `composer require surfnet/stepup-bundle` and use this class from there?

MKodde (approved):

PR #285, opened 2024-05-06
When sorting SPs by name, do so case insentively

thijskh:
So you can find the service "iotroam" near "Internal Gitlab" and not after the "Z".


parijke (approved):

parijke:
Nice @thijskh

PR #286, opened 2024-05-16
Add ARP santizer

parijke:
Added a local function to remove al irrelevant attributes from the ARP returned from the API calls. This way, the API can "grow" without erroring the Arp class

### DIFF
--- a/src/OpenConext/Profile/Tests/fixture/arp-response.json
+++ b/src/OpenConext/Profile/Tests/fixture/arp-response.json
@@ -59,18 +59,13 @@
   ],
   "urn:mace:dir:attribute-def:uid": [
     {
-      "value": "*"
+      "value": "*",
+      "motivation": "my-motivation"
     }
   ],
   "urn:mace:dir:attribute-def:preferredLanguage": [
     {
-      "value": "*"
-    }
-  ],
-  "urn:mace:dir:attribute-def:eduPersonAffiliation": [
-    {
-      "value": "filter-value",
-      "source": "voot",
+      "value": "*",
       "use_as_name_id": false
     }
   ]
--- a/src/OpenConext/Profile/Tests/fixture/arp-response.json
+++ b/src/OpenConext/Profile/Tests/fixture/arp-response.json
@@ -66,5 +66,12 @@
     {
       "value": "*"
     }
+  ],
+  "urn:mace:dir:attribute-def:eduPersonAffiliation": [
+    {
+      "value": "filter-value",
+      "source": "voot",
+      "use_as_name_id": false
+    }
   ]
 }
--- a/src/OpenConext/Profile/Value/Arp.php
+++ b/src/OpenConext/Profile/Value/Arp.php
@@ -40,11 +40,12 @@ final class Arp
         array $arp,
         AttributeDictionary $dictionary = null,
     ): self {
+
+        $arp = self::sanitizeARPData($arp);
+
         // Input validation
         foreach ($arp as $attributeInformation) {
-            if (!is_array($attributeInformation)) {
-                throw new InvalidArpDataException('The attribute information in the arp should be an array.');
-            }
+
             if (!self::isValidAttribute($attributeInformation)) {
                 throw new InvalidArpDataException('The attribute information is formatted invalidly.');
             }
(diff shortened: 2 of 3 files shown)


src/OpenConext/Profile/Value/Arp.php
```diff
@@ -40,11 +40,12 @@ public static function createWith(
...
     ): self {
+
+        $arp = self::sanitizeARPData($arp);
+
         // Input validation
         foreach ($arp as $attributeInformation) {
-            if (!is_array($attributeInformation)) {
-                throw new InvalidArpDataException('The attribute information in the arp should be an array.');
-            }
+
```
parijke, line 48: This if is moved inot the sanitizer

src/OpenConext/Profile/Value/Arp.php
```diff
@@ -186,4 +187,40 @@ public function getMotivationFor(
         }
         return '';
     }
+
+    /**
+     * Accordingly to https://www.pivotaltracker.com/n/projects/1453004/stories/187607790
+     * the ARP data should only contain the following keys: value, source, motivation.
+     * This function will filter out all other keys, which are to be ignored.
+     */
+    private static function sanitizeARPData(array $arp): array
```
MKodde: Nitpick: sanitizeArpData. Abbreviations are considered normal words.

src/OpenConext/Profile/Tests/fixture/arp-response.json
```diff
@@ -66,5 +66,12 @@
     {
       "value": "*"
     }
+  ],
+  "urn:mace:dir:attribute-def:eduPersonAffiliation": [
```
MKodde, line 70: This attribute is already in the fixture, may I suggest you pick a different attr name here?

Also, since we allow the `motivation` field too, that (sh|c)ould be covered in this test too?

src/OpenConext/Profile/Value/Arp.php
MKodde: In the end, you did not move it to the sanitizer right?

src/OpenConext/Profile/Value/Arp.php
```diff
@@ -186,4 +187,40 @@ public function getMotivationFor(
...
+        $sanitizedArp = [];
+        foreach ($arp as $attributeName => $attributeInformation) {
+            if (!is_array($attributeInformation)) {
+                throw new InvalidArpDataException('The attribute information in the arp should be an array.');
+            }
+
+            $sanitizedAttributeInformation = [];
+            foreach ($attributeInformation as $attributeInformationEntry) {
+                if (!is_array($attributeInformationEntry)) {
+                    continue;
```
MKodde, line 209: Is this actually occurring? This feels like a corner case where something bigger is wrong.  Also, given you added this bit of behavior, its only fair to also cover it in test.

src/OpenConext/Profile/Value/Arp.php
```diff
@@ -186,4 +187,40 @@ public function getMotivationFor(
...
+                }
+
+                $sanitizedAttributeInformation[] = array_filter(
+                    $attributeInformationEntry,
+                    function ($key) use ($validKeys) {
+                        return in_array($key, $validKeys);
+                    },
+                    ARRAY_FILTER_USE_KEY
+                );
```
MKodde: :ok_hand:

MKodde (changes requested):
Nice work @parijke. I feel some i's need some dotting.

src/OpenConext/Profile/Value/Arp.php
parijke: @MKodde yes, 206-208

src/OpenConext/Profile/Tests/fixture/arp-response.json
parijke: @MKodde As I have no clue what attributes really are, can you give me a suggestion? Or can it be anything?

src/OpenConext/Profile/Value/Arp.php
parijke: It is, as it was in the create before.... In the test (see ArpTest line 102) it has thios edge...

src/OpenConext/Profile/Tests/fixture/arp-response.json
parijke: And sure, I can add motivation as well (as this wasn't subject of this use case, I did not think about this)

parijke (commented):
@MKodde responed on your spot-on review!

src/OpenConext/Profile/Value/Arp.php
MKodde: Thanks for pointing it out. :+1:

src/OpenConext/Profile/Value/Arp.php
MKodde: The calling part must also be updated now ;) 

TIL: calling methods is not case sensitive :thinking:

src/OpenConext/Profile/Tests/fixture/arp-response.json
MKodde: The attributes are released by the IdP and they contain the data that describe the user. For example there is a mail attribute, a username, commonname, lastname.. You name e'm, there is an attribute for that. And if it does not exist, we can add them to the attribute definition.

I do not know all attributes by memory. I suggest you look in the attribute dictionary file. And pick one from there?

@see: vendor/surfnet/stepup-saml-bundle/src/Resources/config/saml_attributes.yaml

MKodde (commented):
Almost there. Two minor todo's remain

src/OpenConext/Profile/Tests/fixture/arp-response.json
parijke: I added "motivation" to one of the testcases

src/OpenConext/Profile/Value/Arp.php
parijke: Apparantly (didn't knew that either). Will fix though

thijskh (approved):
I've tested this and it indeed solves the problem. Noticed that the newly added test is still failing though.

MKodde (approved):

PR #287, opened 2024-05-17
Set ignore_unknown_attributes to true

parijke:
This value has to be set, otherwise the service is initiated with NULL (which fails typing)

Other possible improvements:
- Config the service with false in the `src/Resources/config/services.yaml` 
- Type the `$ignoreUnkonwnAttributes` variable as bool in the constructor of AttributeDictionary

### DIFF
--- a/config/packages/surfnet_saml.yaml
+++ b/config/packages/surfnet_saml.yaml
@@ -4,6 +4,8 @@ parameters:
 surfnet_saml:
   enable_authentication: true
   hosted:
+    attribute_dictionary:
+      ignore_unknown_attributes: true
     service_provider:
       enabled: true
       assertion_consumer_route: profile.saml_consume_assertion


config/packages/surfnet_saml.yaml
```diff
@@ -4,6 +4,8 @@ parameters:
 surfnet_saml:
   enable_authentication: true
   hosted:
+    attribute_dictionary:
+      ignore_unknown_attributes: true
```
MKodde, line 8: If this fixes the problem then that's a step in the right direction.

I do think the old behavior was that unknown attributes were *not* ignored as that is the default behavior of the saml bundle.

1. Might be better to fix this issue at the source. In other words, allow the property to not be configured, and default it to `false`?

MKodde (commented):
I looked into the SAML bundle implementation for ignoring unknown attributes. And I think you are moving in the wrong direction with this change. See my comment below.

config/packages/surfnet_saml.yaml
parijke: Yes, that is fine by me. Would you like me to make a PR for that?

thijskh: Yes please

parijke:
Is not needed anymore

thijskh:
So we still need this to restore Profile's previous behaviour of accepting unknown attributes

thijskh (approved):

PR #297, opened 2024-08-28
Added spacing in inviteRoles__top to avoid overlap

oharsta:


MKodde:
Thanks for adding the `gap` @oharsta, This helps, but does not completely solve the overlap (at least not as I observed it). I still had to set a max-width and added an overflow-hidden on the img container

assets/css/pages/inviteRoles/inviteRoles.scss
```diff
@@ -28,12 +29,14 @@
...
-    height: 90px;
-    min-width: 90px;
     align-content: flex-start;
+    overflow: hidden;
+    max-width: 120px;
+    min-width: 120px;
     margin-right: 1rem;
     img {
       height: inherit;
+      min-width: inherit;
```
thijskh, line 39: This should likely be `max-width`?

thijskh (approved):

